PR #56, opened 2021-07-06
46: Search feature

oneroyalace:
Adds a search functionality and interface for querying Instagram and Twitter user/content tables. Currently works by issuing individual select queries to each table, an implementation that's not pretty and may be annoying to  maintain if we want to search by more features in the future.  

This PR needs some UI work to properly display profiles in search results.  

I think there's a much cleaner way to get the whole thing working using [PgSearch](https://github.com/Casecommons/pg_search), which I'll dig into later this week


app/models/instagram_post.rb
```diff
@@ -2,6 +2,8 @@
 
 class InstagramPost < ApplicationRecord
   include ArchivableItem
+  include PgSearch::Model
```
oneroyalace, line 5: These will only be necessary if we decide to go with a `PgSearch` implementation

oneroyalace:
Some to-dos: 

- [ ] Write a search results preview erb that takes image_data/video_data from UnifiedPosts attributes
- [ ] Look into linking the two views so data doesn't need to be duplicated across them
- [ ] Create a search results erb for profiles (I'm currently just printing their display name in plain text)

app/models/unified_post.rb
```diff
@@ -0,0 +1,9 @@
+class UnifiedPost < ApplicationRecord
+  def readonly?
+    true
+  end
+
```
oneroyalace, line 5: @cguess To avoid data duplication, I could put a method here that calls `Post.where(display_name ~* '#{search_term}'` and then joins it with the Users table on author id. I'd have to check whether that caches properly

app/models/unified_post.rb
```diff
@@ -0,0 +1,26 @@
...
+  self.primary_key = :post_id  # think I need this for the query generation.
+
+  # pg_search_scope :search_by_text, against: :text
+  pg_search_scope(
+    :search_,
+    against: :tsv_document,
+    using: {
+      tsearch: {
+        # dictionary: 'english'  #TODO: Uncomment line to enable stemming
```
oneroyalace, line 12: By default, `postgres` doesn't stem search terms. I believe we can enable that option by setting `dictionary: 'english'`

app/models/unified_user.rb
```diff
@@ -0,0 +1,23 @@
...
+  include PgSearch::Model
+  self.primary_key = :author_id
+
+  pg_search_scope(
+    :search_users,
+    against: :tsv_document,
+    using: {
+      tsearch: {
+        tsvector_column: 'tsv_document',
+        # prefix: true  # do we really want this for user searches?
```
oneroyalace, line 11: If this is enabled, searching for 'Bid' will bring up users named 'Biden'. Not sure whether we want that

app/models/unified_user.rb
```diff
@@ -0,0 +1,23 @@
...
+      }
+    }
+  )
+
+  def readonly?
+    true
+  end
+
+  def self.refresh
+    Scenic.database.refresh_materialized_view(table_name, concurrently: true, cascade: false)
```
oneroyalace, line 21: Allowing concurrent refreshes keeps the view from being locked up

PR #58, opened 2021-07-19
Update rbenv installation instructions

oneroyalace:
When using rbenv, installing ruby versions with the `--keep` or `-k` flag ensures that Ruby source files are locally available for gems that need them, like `dhash-vips`

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,9 @@ Follow the instructions in any of the repos to properly install it. Make sure to
 
 ##### [Rbenv](https://github.com/rbenv/rbenv)
 *This is the one I (@cguess) use.*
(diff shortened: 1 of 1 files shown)


cguess (commented):
Can you add a quick explanation why? It's pretty uncommon so best to explain for someone new coming across this.

cguess (approved):

Nakilon:
> Can you add a quick explanation why? It's pretty uncommon so best to explain for someone new coming across this.

Hello.
The explanation is that the C extension here https://github.com/Nakilon/dhash-vips/blob/master/idhash.c uses constants and functions that aren't defined in basic header files within usual Ruby installation package. They are in the `bignum.c` that is only can be found in full Ruby source code. I don't know why. I guess no one cared because no one wrote native extensions around Bignums.

However! The gem should be able to be installed and work without the native extensions compiled -- it will just work slower because of the fallback to pure Ruby `distance` implementation.

PR #59, opened 2021-07-19
Add image search

cguess:
This PR adds reverse image searching It is very much not ready for full-scale production, but good enough for now.

It also reorganizes the model folder, while insuring backwards capability for ease of use.

There's a few things to do after this is merged though
- [ ] The results page is not quite done, there's links that need to be created
- [ ] There's no individual entry page yet
- [ ] We have filler data for media review


oneroyalace:
I'm seeing `uninitialized constant MediaModels::Images::InstagramImage::InstagramPost` inside the  `search_result_list_item` view on this line 
https://github.com/TechAndCheck/zenodotus/blob/f28b60e39c50c24b2a23ec3cb8ee248cbd638535/app/views/image_search/_search_result_list_item.html.erb#L18-L18

For some reason, Rails doesn't seem to be pulling the associated post in.

PR #82, opened 2021-08-27
Run materialized view refreshes using ActiveJob

oneroyalace:
This PR moves materialized view refreshes into an ActiveJob


cguess (changes requested):
This test is failing. It seems as if `/app/model/sources/tweet.rb` doesn't actually call the new function?

oneroyalace:
> This test is failing. It seems as if `/app/model/sources/tweet.rb` doesn't actually call the new function? Did something not get committed?

This turned out to be an ActiveJob issue. To execute ActiveJobs in tests, you need to explicitly run `perform_enqueued_jobs`. Since we're running a test here that's creates a job from a controller and doesn't inherit from `ActiveJob::TestCase`, we have to include `ActiveJob::TestHelper` to get access to that method.

PR #84, opened 2021-08-31
Switch from using Zorki locally to external host

cguess:
Due to the fact that Instagram (and probably everyone else) blocks AWS,
Azure, etc. IP addresses we need to run the scraping from somewhere
else. To that end I created a server for it and am running it locally in
my apartment off a Raspberry Pi 3. So this changes the behaviour to call
the server instead of the local version of the gem.

This adds two new required environment variables:
`ZORKI_SERVER_URL` which points to the right server
`ZORKI_AUTH_KEY` which is the auth key for authenticating the requests

Everything else stays exactly the same.

To test:
1. Pull this branch
2. Run `bundle install` since the zorki gem is now removed from the
   requirements
3. Add the two environment variable however you're handling that
4. Run `rails t` and it'll pass

### DIFF
--- a/app/models/sources/instagram_user.rb
+++ b/app/models/sources/instagram_user.rb
@@ -13,11 +13,11 @@ class Sources::InstagramUser < ApplicationRecord
   # @params birdsong_users [Array[Zorki:User]] an array of tweets grabbed from Zorki
   # @returns [Array[ArchiveEntity]] an array of +ArchiveEntity+ with type of InstagramUser that have been
   #   saved
-  sig { params(zorki_users: T::Array[Zorki::User]).returns(T::Array[ArchiveEntity]) }
+  sig { params(zorki_users: T::Array[Hash]).returns(T::Array[ArchiveEntity]) }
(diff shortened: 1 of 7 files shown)


app/models/sources/instagram_user.rb
```diff
@@ -55,19 +55,31 @@ def service_id
...
-      number_of_posts:     zorki_user.number_of_posts,
-      followers_count:     zorki_user.number_of_followers,
-      following_count:     zorki_user.number_of_following,
-      verified:            zorki_user.verified,
-      profile:             zorki_user.profile,
-      url:                 zorki_user.profile_link,
-      profile_image_url:   zorki_user.profile_image_url,
-      profile_image:       File.open(zorki_user.profile_image, "rb"),
+    # We create a temp file and write the image data to it, which yea, is dumb,
+    # and there may be a better way to do it, but this works to fix the encoding issues
```
oneroyalace, line 61: 👍🏿

oneroyalace (approved):
LGTM. I'm able to hit the pi scraper locally

Issue #86, opened 2021-09-13
Add MediaReview to data model


oneroyalace:
Each `ArchiveItem` should have one MediaReview. For now, we'll just create a minimal set of attributes for MediaReview (maybe just url?)

PR #87, opened 2021-09-13
Add mediareview model, relations, and db table

oneroyalace:
This PR adds a basic `MediaReview` model with a few attributes. A `MediaReview` belongs to an `ArchiveItem`, and will eventually be created in the IngestController here: 

https://github.com/TechAndCheck/zenodotus/blob/8d405d89299146dc06c1b9d20a81a1e26e7ada13/app/models/archive_item.rb#L16-L31

### DIFF
--- a/db/migrate/20210913200056_create_media_review.rb
+++ b/db/migrate/20210913200056_create_media_review.rb
@@ -2,9 +2,9 @@ class CreateMediaReview < ActiveRecord::Migration[6.1]
   def change
     create_table :media_review, id: :uuid do |t|
       t.timestamps
-      t.text :link, null: false
-      t.text :authenticity, null: false
-      t.text :context, null: false
+      t.text :original_media_link, null: false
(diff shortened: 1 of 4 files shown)


db/migrate/20210913200056_create_media_review.rb
```diff
@@ -0,0 +1,12 @@
+class CreateMediaReview < ActiveRecord::Migration[6.1]
+  def change
+    create_table :media_review, id: :uuid do |t|
+      t.timestamps
+      t.text :media_link, null: false
```
cguess, line 5: We should maybe rename these? If they all start with `media` does it become a bit repetitious?

oneroyalace: You're right. I'll cut the prefix

cguess (approved):

oneroyalace:
closes #86

Issue #85, opened 2021-09-13
Conditional rendering


oneroyalace:
Implement conditional rendering based on user attributes

oneroyalace:
Closed by #89

PR #89, opened 2021-09-14
Conditionally render views based on user attributes

oneroyalace:
This PR introduces a user attribute, `restricted`, which it uses to conditionally filter certain data out of views. Some of that data will is replaced by attributes from the newly introduced `MediaReview` model, which has a 1:1 relationship with `ArchiveItem`s.


oneroyalace:
When merged, closes #85

app/controllers/archive_controller.rb
```diff
@@ -3,20 +3,17 @@
 class ArchiveController < ApplicationController
   # It's the index, list all the archived items
 
-  # before_action :authenticate_user!
```
cguess, line 6: Why is all this removed? we still want users to log in.

oneroyalace: Ah, whoops. It was commented out, so I figured we weren't requiring authentication for the `ArchiveController`. Should I reinstate/uncomment it?

cguess (approved):

PR #92, opened 2021-09-20
Archive embedded MediaReview via URL 

### LINKED ISSUE #91: Add ability to archive material from sites using the MediaReview/ClaimReview schema
Right now, we have a hook that lets us archive media if we get a direct link to it. Many fact-checkers though, implement ClaimReview (and will implement MediaReview) by using tools  that embed JSON into fact check pages. (see e.g. [FactStream](https://fullfact.org/economy/wine-regulation-cost/)). 

We need to add a hook that accepts URLs, scrapes them to check for the presence of ClaimReview/MediaReview, then after making sure that the pertinent JSON object is well-formed, archives the media they point to (and links it in our data model to its associated `MediaReview` data).

oneroyalace:
This PR enables API users to add to media/ClaimReview objects to the archive by submitting a link to a page in which well-formed MediaReview objects are embedded. 

Currently, I'm testing this by passing in a link to a GithHub page hosted out of one of my repos. At some point, we should probably mock that page out (and hey, unrelated, maybe mock Zorki, too). 

# Testing
`rails t` 

# Requirements
No migration required

closes #91

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -14,6 +14,7 @@ protected
 
   sig { void }
   def authenticate_user_from_api_key!
+    # return true
     if params[:api_key].blank?
       render json: {
         error: "Unauthorized credentials, please check your API Key"
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -14,7 +14,6 @@ protected
 
   sig { void }
   def authenticate_user_from_api_key!
-    # return true
     if params[:api_key].blank?
       render json: {
         error: "Unauthorized credentials, please check your API Key"
--- a/app/controllers/archive_controller.rb
+++ b/app/controllers/archive_controller.rb
@@ -32,10 +32,6 @@ class ArchiveController < ApplicationController
   # @params {url_to_archive} the url to pull in
   sig { void }
   def submit_url
-    # unless user_signed_in?
-    #   format.html { redirect_to :new_user_session_path }
-    #   return
-    # end
     typed_params = TypedParams[SubmitUrlParams].new.extract!(params)
     url = typed_params.url_to_archive
     object_model = ArchiveItem.model_for_url(url)
--- a/app/controllers/ingest_controller.rb
+++ b/app/controllers/ingest_controller.rb
@@ -59,39 +59,124 @@ class IngestController < ApplicationController
     const :media_review_json, String
   end
 
-  # Submit MediaReview data, which the URL will be scraped from
+  # A class representing the allowed params into the `submit_media_review_source` endpoint
+  class SubmitMediaReviewSourceParams < T::Struct
+    const :url, String
+  end
+
+
+  # Creates a MediaReview and ArchiveItem based on a POSTed MediaReview JSON object
+  #
+  #  @params {media_review_json} A MediaReview JSON object
   sig { void }
   def submit_media_review
-    # byebug
     # TODO: Spin off an active job to handle this
     typed_params = TypedParams[SubmitMediaReviewParams].new.extract!(params)
-    media_review_json = JSON.parse(typed_params.media_review_json)
+    media_review_json = JSON.parse typed_params.media_review_json
+    response_payload = archive_from_media_review(media_review_json)
 
+    render(json: response_payload, status: response_payload.has_key?(:error) ? 400 : 200)
+
+  rescue JSON::ParserError
+    response_payload = {
+      error_code: ApiErrors::JSONParseError.code,
+      error: ApiErrors::JSONParseError.message,
+      failures: media_review_json
+    }
+    render(json: response_payload, status: 400)
+  end
+
+
+  # Creates MediaReview and ArchiveItem(s) based on MediaReview founded at the page pointed to by the URL param
+  #
+  # @params {url} the url to look at
+  sig { void }
+  def submit_media_review_source
+    typed_params = TypedParams[SubmitMediaReviewSourceParams].new.extract!(params)
+    mediareview_array = find_media_review_in_page typed_params.url
+
+    unless mediareview_array.length > 0
+      failure_response = {
+        response_code: 40,
+        response: "Could not find MediaReview in webpage"
+      }
+      render(json: failure_response, status: 400)
+      return
+    end
+
+    # Archive items
+    responses = mediareview_array.each do |review|
+      archive_from_media_review review
+    end
+
+    archive_success = responses.all? { |response| !response.has_key?(:error)}
+    if archive_success
+      success_response = {
+        response_code: ApiResponseCodes::Success.code,
+        response: "Successfully archived #{responses.length} MediaReview object(s) and associated media"
+      }
+      render(json: success_response, status:200 )
+      return
+    end
+
+    failure_response = {
+      response_code: 40,
+      response: "Failed to archive one or more MediaReview items",
+      information: responses
+    }
+    render(json: failure_response, status: 400)
+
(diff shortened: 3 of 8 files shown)


oneroyalace:
https://github.com/TechAndCheck/zenodotus/blob/a26b9a1652da6adf1091b5a0ef330f65de95570c/app/controllers/ingest_controller.rb#L108-L111

There may be idempotency issues with this if one of the `ArchiveItem` creations fails.

app/controllers/application_controller.rb
```diff
@@ -14,6 +14,7 @@ def json_request?
 
   sig { void }
   def authenticate_user_from_api_key!
+    # return true
```
cguess, line 17: Redundant comment

app/controllers/archive_controller.rb
```diff
@@ -32,10 +32,6 @@ class SubmitUrlParams < T::Struct
   # @params {url_to_archive} the url to pull in
   sig { void }
   def submit_url
-    # unless user_signed_in?
```
cguess, line 35: Removing this, was it placed elsewhere?

app/controllers/ingest_controller.rb
```diff
@@ -59,46 +59,131 @@ class SubmitMediaReviewParams < T::Struct
...
+  # Creates a MediaReview and ArchiveItem based on a POSTed MediaReview JSON object
+  #
+  #  @params {media_review_json} A MediaReview JSON object
   sig { void }
   def submit_media_review
-    # byebug
     # TODO: Spin off an active job to handle this
     typed_params = TypedParams[SubmitMediaReviewParams].new.extract!(params)
-    media_review_json = JSON.parse(typed_params.media_review_json)
+    media_review_json = JSON.parse typed_params.media_review_json
```
cguess, line 75: Let's stay consistent (I know I'm not the best at this either) with paraenthatsis, and require them for surrounding parameters.

app/controllers/ingest_controller.rb
```diff
@@ -59,46 +59,131 @@ class SubmitMediaReviewParams < T::Struct
...
+      error: ApiErrors::JSONParseError.message,
+      failures: media_review_json
+    }
+    render(json: response_payload, status: 400)
+  end
+
+
+  # Creates MediaReview and ArchiveItem(s) based on MediaReview founded at the page pointed to by the URL param
+  #
+  # @params {url} the url to look at
```
cguess, line 92: Make sure the comment match up with the function. So this should have no params

app/controllers/ingest_controller.rb
```diff
@@ -59,46 +59,131 @@ class SubmitMediaReviewParams < T::Struct
...
+  sig { void }
+  def submit_media_review_source
+    typed_params = TypedParams[SubmitMediaReviewSourceParams].new.extract!(params)
+    mediareview_array = find_media_review_in_page typed_params.url
```
cguess, line 96: Parentheses.

app/controllers/ingest_controller.rb
```diff
@@ -59,46 +59,131 @@ class SubmitMediaReviewParams < T::Struct
...
+    mediareview_javascript = /<script.*?>(\[.*MediaReview.*\]).*<\/script>/
+
+    response = Typhoeus.get url
+    if response.response_code != 200
+      return []
+    end
+    body = response.body
+
+    mediareview_string = body.match mediareview_javascript
+    unless mediareview_string  # catch pages that don't have embedded MediaReview
```
cguess, line 151: you can refactor this a bit nicer to `return [] unless mediareview_string` all on one line. Ruby is a tad weird.

app/controllers/ingest_controller.rb
```diff
@@ -59,46 +59,131 @@ class SubmitMediaReviewParams < T::Struct
...
+    mediareview_array = mediareview_string.captures.first
+    JSON.parse mediareview_array
+  end
+
+  # Creates ArchiveItems (and MediaReview objects) based on a MediaReview JSON object
+  # {media_review_json}: A MediaReview JSON object
+  def archive_from_media_review(media_review_json)
     unless validate_media_review(media_review_json)
-      error = {
+      response_payload = {
```
cguess, line 163: I'd refactor this to a single line. Something like
```
return { 
        error_code: ApiErrors::JSONValidationError.code,
         error: ApiErrors::JSONValidationError.message,
         failures: media_review_json }
} unless validate_media_review(media_review_json)

File: app/controllers/ingest_controller.rb [hunk 8]
```diff
@@ -59,46 +59,131 @@ class SubmitMediaReviewParams < T::Struct
...
     }
-
-    render json: response, status: 200
-  rescue JSON::ParserError
-    error = {
-      error_code: ApiErrors::JSONParseError.code,
-      error: ApiErrors::JSONParseError.message,
-      information: "" }
-    render(json: error, status: 400)
+    response_payload
```
cguess, line 176: no need to declare a variable just to return it, just make the hash the last element for auto-returns

app/models/archive_item.rb
```diff
@@ -25,7 +25,14 @@ def self.create_from_media_review(media_review)
...
     # However, we'll refactor another time after it works
-    object.update!({ media_review: media_review })
+
+    MediaReview.create(
+      original_media_link: url,
+      media_authenticity_category: media_review["mediaAuthenticityCategory"],
+      original_media_context_description: media_review["originalMediaContextDescription"],
+      archive_item_id: object.id
+    )
+    # object.update!({ media_review: media_review })
```
cguess, line 35: Redundant

test/controllers/ingest_controller_test.rb
```diff
@@ -2,19 +2,20 @@
 
 require "test_helper"
 
+# rubocop:disable ClassLength
```
cguess, line 5: Should we just remove this cop in general?

cguess (changes requested):

app/controllers/application_controller.rb
oneroyalace: Oops, threw this in while testing. I'll remove it

app/controllers/archive_controller.rb
oneroyalace: I believe it's redundant. Earlier in this file, we have:

https://github.com/TechAndCheck/zenodotus/blob/a26b9a1652da6adf1091b5a0ef330f65de95570c/app/controllers/archive_controller.rb#L6-L6

app/controllers/ingest_controller.rb
oneroyalace: Sure thing. I got a little drunk on syntactical freedom

app/controllers/ingest_controller.rb
oneroyalace: Is there a way to indicate that it has access to params not in the function definition?

app/controllers/ingest_controller.rb
oneroyalace: Ahh I was worried that that statement would return null. Neat

test/controllers/ingest_controller_test.rb
oneroyalace: Sounds good to me. I'll change the config file

app/controllers/archive_controller.rb
cguess: Good call

app/controllers/ingest_controller.rb
cguess: No, since these are just signatures. Scope should imply access ability.

cguess (approved):

PR #97, opened 2021-10-07
Log the user id of searchers

oneroyalace:
This PR ensures that a `user_id` is recorded when a search is made. The attribute was previously always null.

### DIFF
--- a/app/controllers/text_search_controller.rb
+++ b/app/controllers/text_search_controller.rb
@@ -1,5 +1,7 @@
 # typed: ignore
 class TextSearchController < ApplicationController
+  before_action :authenticate_user!
+
   sig { void }
   def index
     @search = TextSearch.new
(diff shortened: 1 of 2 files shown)


app/controllers/text_search_controller.rb
```diff
@@ -18,7 +20,7 @@ class SubmitUrlParams < T::Struct
   def search
     typed_params = TypedParams[SubmitUrlParams].new.extract!(params)
     # Create a search object
-    search = TextSearch.create(query: typed_params.query)
+    search = TextSearch.create(query: typed_params.query, user_id: current_user.id)
```
cguess, line 23: Looking through this, I think we should probably add a `belongs_to` method to `TextSearch`. This would allow just `TextSearch.create(query: typed_params.query, user: current_user)`

We could also even add a `before_create` to the `TextSearch` model to automatically assign the current user to any model being created.

oneroyalace: Can do!

PR #101, opened 2021-10-11
Data export feature

### LINKED ISSUE #98: Feature: Allow users to export MediaReview data
Ideally, we'd allow users to export the full dataset (from the archive page)  or subsets of the data (from a search/user page).

oneroyalace:
This PR creates a feature that lets users export data from the archive, either from the archive home page (exporting details of every item in the full archive) or from a social media user page (exporting details on only the media authored by that user). 

closes #98


oneroyalace:
Included in #109

PR #131, opened 2022-01-21
Add method-level type checking to controllers + models 

oneroyalace:
This PR adds Sorbet method signatures to models and controllers. It also enables type checking for them (!)

## Test
- rails t


cguess:
I'm getting the following error when testing:
```
Error:
ArchiveItemTest#test_can_generate_ArchiveItem_JSON_export:
NoMethodError: private method `Array' called for T:Module
    app/models/archive_item.rb:41:in `block in <class:ArchiveItem>'
    test/models/archive_item_test.rb:8:in `block in <class:ArchiveItemTest>'
```

Can you take a look?

oneroyalace:
Test fixed @cguess. I referenced. I tried to access sorbet's array type with `T.Array` instead of `T:Array`

PR #134, opened 2022-01-29
Pre-compute tsvectors to speed up postgres text search 

oneroyalace:
This PR adds a stored generated column to the `pg_search_documents` table containing `tsvector`s computed from document's searchable content. The table should look something like the one below and doesn't require manual updating. 
```
+----+----------------------+-----------------------------+-----+-----------------------+
| id |   searchable_type    |           content           | ... |   content_tsvector    |
+----+----------------------+-----------------------------+-----+-----------------------+
|  0 | Sources::TwitterUser |  kairyssdal                 | ... | 'kairyssd':1          |
|  1 | Sources::Tweet       |  This is an automatic mango | ... | 'automat':4 'mango':5 |
+----+----------------------+-----------------------------+-----+-----------------------+
```

To test:
1. `rails db:migrate`
2. `rails t`

### DIFF
--- a/app/models/sources/twitter_user.rb
+++ b/app/models/sources/twitter_user.rb
@@ -6,7 +6,7 @@ class Sources::TwitterUser < ApplicationRecord
   include PgSearch::Model
 
 
-  multisearchable against: :handle
+  multisearchable against: [:handle, :display_name]
 
   # The tweets that a TwitterUser have authored
   has_many :tweets, foreign_key: :author_id, dependent: :destroy
--- a/app/models/sources/twitter_user.rb
+++ b/app/models/sources/twitter_user.rb
@@ -6,12 +6,7 @@ class Sources::TwitterUser < ApplicationRecord
   include PgSearch::Model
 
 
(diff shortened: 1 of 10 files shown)


app/models/sources/twitter_user.rb
```diff
@@ -5,6 +5,7 @@ class Sources::TwitterUser < ApplicationRecord
   include ImageUploader::Attachment(:profile_image) # adds an `image` virtual attribute
   include PgSearch::Model
 
+
```
cguess, line 8: Remove extra whitespace.

config/initializers/pg_search.rb
```diff
@@ -0,0 +1,8 @@
+PgSearch.multisearch_options = {
+  using: {
+    tsearch: {
+      dictionary: "english",
```
cguess, line 4: Just making a note that we may want to look into multilingual models here for my future reference. English is fine for now.

cguess (approved):
one style change, but otherwise you're good to merge

PR #141, opened 2022-02-17
135 scraping callback

cguess:
This finishes up the scraper callback system as well as all the tests to
make sure this works properly in a CI pipeline.

Specifically this adds a workflow allowing Hypatia to call back to the Zenodotus instance when a scrape is over. This allows scrapes to be run asynchronously so as to not break the scrapers (since you can only run one scrape at a time).

To test:
1. `rails db:migrate`
2. Ensure that Hypatia is running locally on port 3001
3. `rails t`. There should be only 11 errors (all facebook related)


app/models/scrape.rb
```diff
@@ -0,0 +1,18 @@
...
+    }, _prefix: true
+
+  belongs_to :archive_item, optional: true
+
+  def fulfill(response)
+    archive_item = case self.scrape_type
+                   when "instagram"
+                     Sources::InstagramPost.create_from_zorki_hash([response])
+                   when "facebook"
+                     Sources::InstagramPost.create_from_forki_hash([response])
```
cguess, line 13: Whoops typo

app/models/scrape.rb
```diff
@@ -0,0 +1,18 @@
+class Scrape < ApplicationRecord
+  enum scrape_type: {
+      twitter: "twitter", instagram: "instagram", facebook: "facebook"
+    }, _prefix: true
+
+  belongs_to :archive_item, optional: true
```
cguess, line 6: Switch to has_one

Issue #156, opened 2022-03-28
Update README with latest instructions


reefdog:
I just went through the (excellent) README, and found it (1) tremendously useful and (2) largely up to date. There were a few items that @cguess and I discovered were no longer the case. I'm going to make a few changes so it's ready to run out of the box.

PR #157, opened 2022-03-28
Add ability to manage scrapes

cguess:
This adds the ability to view, resubmit and delete scrapes that have not
been completed yet.

To test:
1. `rails db:migrate`
1. `rails test`

### DIFF
--- a/app/views/jobs_status/index.html.erb
+++ b/app/views/jobs_status/index.html.erb
@@ -29,7 +29,7 @@
       <th class="border text-center">Scrape Type</th>
       <th class="border text-center">Enqueued at</th>
       <th class="border text-center">Url to scrape</th>
-      <th class="border text-center"></th>
+      <th class="border text-center" colspan="2">Actions</th>
     </tr>
     </thead>
     <tbody id="jobs-table-body">
--- a/app/views/jobs_status/index.html.erb
+++ b/app/views/jobs_status/index.html.erb
@@ -41,11 +41,11 @@
           <td><%= link_to scrape.url, scrape.url %></td>
           <td><%= link_to "Delete", job_status_delete_path(scrape.id), data: {
                     turbo_method: :delete,
-                    turbo_confirm: "Are you sure?"
+                    turbo_confirm: "Are you sure you would like to stop this scrape?"
                   } %></td>
           <td><%= link_to "Resubmit", job_status_resubmit_path(scrape.id), data: {
                     turbo_method: :post,
-                    turbo_confirm: "Are you sure?"
+                    turbo_confirm: "Are you sure you would like to resubmit this scrape?"
                   } %></td>
         </tr>
       <% end %>
--- a/app/views/jobs_status/index.html.erb
+++ b/app/views/jobs_status/index.html.erb
@@ -1,4 +1,5 @@
-<div id="jobs-div">
+<div class="text-3xl font-bold">Active Jobs</div>
+<div id="jobs-div" class="mt-5">
   <table id="jobs-table" class="table-auto w-full">
     <thead id="jobs-table-head">
     <tr>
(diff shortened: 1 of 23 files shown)


config/tailwind.config.js
```diff
@@ -0,0 +1,21 @@
...
+module.exports = {
+  content: [
+    './app/helpers/**/*.rb',
+    './app/javascript/**/*.js',
+    './app/views/**/*.{erb,haml,html,slim}'
+  ],
+  theme: {
+    extend: {
+      fontFamily: {
+        sans: ['Inter var', ...defaultTheme.fontFamily.sans],
```
reefdog, line 12: Just to sanity-check `Inter var` is the valid name of the font?

app/views/jobs_status/index.html.erb
```diff
@@ -18,3 +19,36 @@
...
+    <tbody id="jobs-table-body">
+      <% @scrapes.each do |scrape| %>
+        <tr class="border text-gray-600 text-center">
+          <td><%= scrape.id %></td>
+          <td><%= scrape.scrape_type %></td>
+          <td><%= scrape.created_at %></td>
+          <td><%= link_to scrape.url, scrape.url %></td>
+          <td><%= link_to "Delete", job_status_delete_path(scrape.id), data: {
+                    turbo_method: :delete,
+                    turbo_confirm: "Are you sure?"
```
reefdog, line 44: UX request: can you customize this confirmation so it's more clear to the user which action they're confirming? Suggested: "Are you sure you want to delete this job?"

app/views/jobs_status/index.html.erb
```diff
@@ -18,3 +19,36 @@
...
+                  } %></td>
+          <td><%= link_to "Resubmit", job_status_resubmit_path(scrape.id), data: {
+                    turbo_method: :post,
+                    turbo_confirm: "Are you sure?"
```
reefdog, line 48: UX request: can you customize this confirmation so it's more clear to the user which action they're confirming? Suggested: "Are you sure you want to resubmit this job?"

reefdog (changes requested):
Very small UX request around confirmation language.

Also, when I run this I get a bunch of failing tests, but I suspect this is due to an incomplete setup. I'll resolve that with you separately (and update the README for #156 further as needed), and then check this again once my setup actually works.

config/tailwind.config.js
reefdog: Answering my own question here: [yup, `Inter var` is correct](https://tailwindui.com/documentation#optional-add-the-inter-font-family).

reefdog: Answering my own question here: [yup, `Inter var` is correct](https://tailwindui.com/documentation#optional-add-the-inter-font-family).

cguess:
Yes. I’ll get to it in the next hour

-Christopher Guess
***@***.***
US/WhatsApp/Signal: +1 262.893.1037
PGP: AAE7 5171 0D81 B45B – https://keybase.io/cguess

cguess:
All cleaned up. @reefdog please take a look

reefdog:
@cguess Tests are currently running, but I just fired up the app and was greeted with a `NameError in Archive#index`, I think because of the same `jobs_status_path` issues you were fixing in 2db1b8da1ed4944a719b41cdb5682f71699e9be7. Might need a few more snipes:

image

reefdog (changes requested):
@cguess It's a clean test run! 🎉 The only issue is the error on `jobs_status_path` that I noted above, that causes rendering the header to throw an error. After that I think this is good to go.

…although it strikes me that I don't actually know how to test the _functionality_ of this in the browser.

reefdog (approved):

PR #163, opened 2022-03-30
Fix a bunch of type errors between Hypatia

cguess:
This fixes a few little bugs that caused some broken tests when scraping
primarily facebook and twitter. Why this happened when we though we had
fixed it awhile ago? Unsure.

However, all tests should fully pass after this.

To test:
`rails t`

### DIFF
--- a/app/media_sources/twitter_media_source.rb
+++ b/app/media_sources/twitter_media_source.rb
@@ -11,6 +11,17 @@ class TwitterMediaSource < MediaSource
     ["www.twitter.com", "twitter.com"]
   end
 
+  # This is a flag if the scraper is run by Hypatia or locally, mostly so the ScraperJob can know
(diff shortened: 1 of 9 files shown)


app/media_sources/twitter_media_source.rb
```diff
@@ -11,6 +11,17 @@ def self.valid_host_name
     ["www.twitter.com", "twitter.com"]
   end
 
+  # This is a flag if the scraper is run by Hypatia or locally, mostly so the ScraperJob can know
+  # Default is false if not implemented
```
reefdog, line 15: I would've left off this bit of boilerplate since it's only relevant to the default method definition, but nbd.

reefdog (approved):
Huzzah! Thank you for tackling this so we have a clean test run.

PR #167, opened 2022-04-01
Fix type issue when extracting from facebook

cguess:
When not forcing a request to a `FacebookMediaSource` we had a bug in
the typing that would mean it would always error out. Whoops. There
didn't seem to be a test for this either. I've fixed both.

To test
`rails t test/media_sources/facebook_media_source_test.rb`


oneroyalace (approved):
LGTM

PR #170, opened 2022-04-04
Add lint job to ci

cguess:


reefdog (approved):
It's tough to review this since so many old commits came along for the ride. We should keep on eye on whether this keeps happening, because it'll make reviews rough!

Otherwise, hey, this seems to work! 🎉

My only request/suggestion in the future is to feel comfortable amending + force-pushing for things like 8dd13e19d052f1e214632ae6da24e89a7afc1401 since there's no value to the discrete commit.

PR #172, opened 2022-04-06
Prefix Active Job queues with Zenodotus namespace

### LINKED ISSUE #171: Need to prefix queue names when running on shared machines
When running in the same, uh, Redis space as another Sidekiq instance (such as running Zenodotus and Hypatia simultaneously on the same local machine), the jobs from each app get mixed up if not running on separate queues. This results in such errors as the `ScraperJob` class (from Zenodotus) being referenced in the Hypatia headspace (I'm a professional programmer) and Sidekiq throwing an error because it doesn't know what `ScraperJob` is but it wonders if you mean `ScrapeJob`, etc.

Solution: configure development.rb and test.rb to prefix Active Job queue names uniquely!

_This has [Hypatia companion issue #20](https://github.com/TechAndCheck/hypatia/issues/20) because the same problem and solution apply there._

reefdog:
This PR prefixes Active Job queue names with `zenodotus_{environment}` to avoid queue collisions when Zenodotus is being run on the same machine as another Sidekiq-using app (like Hypatia locally).

To test:
- [ ] Clean `rails test` (except the existing test failure on `master`)
- [ ] Run the app and make sure archiving URLs still works
- [ ] Bonus points for doing the above while also running Hypatia

🚨 Pay particular attention to the fact that I enabled prefixes in production, too. You may not want this! But I figured, hey, why not.

Closes #171

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -68,5 +68,8 @@ Rails.application.configure do
   # Uncomment if you wish to allow Action Cable access from any origin.
   # config.action_cable.disable_request_forgery_protection = true
 
+  # Prefix job queues names to avoid collisions
+  config.active_job.queue_name_prefix = "zenodotus_development"
+
   config.hosts << "showoff-reporterslab.pagekite.me"
 end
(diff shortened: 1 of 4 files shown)


cguess (approved):
LGTM

oneroyalace (approved):
lgtm

reefdog:
Thanks both!

PR #174, opened 2022-04-08
Add youtube cop

cguess:
Add custom YouTube style cop
Due to namespacing issues we have to stylize `YouTube` as `Youtube`.
This adds a custom Rubocop cop to enforce this.

To test:
In a random file add something like `YouTubeStuff.do_something` and then
run `rubocop` on the command line. It should flag the name. Remove the
line and run again and it should be gone.`


cguess:
Note: This did the weird "bring a ton of old commits along" thing. I"m going to fix it, but I didn't realize it was happening until after the PR was created. There's only one file added here that's relevant, so please try the rubocop cop and I'm going to fix this using a different branch with nothing of consequence on it.

reefdog:
The `rubocop` rule works as designed! 👏

As soon as we get this commit history business sorted, I'm good with merging.

PR #179, opened 2022-04-14
Equalize force nonforce

cguess:
This PR equalizes the style of response from Hypatia when forcing vs running scrape async. Confirm the Hypatia PR before this one.

Nothing functionally should have changed. This also speeds up tests a bit though by caching responses from Hypatia when running the whole thing.

To test:
- `bundle install`
- `rails t`


cguess:
There's a lot of commits here so I wouldn't mind merging, some of them came in from another, since abandoned branch.

.gitconfig
```diff
@@ -0,0 +1,5 @@
+[alias]
```
oneroyalace, line 1: Are these documented anywhere? I'm not so great at reading shell

reefdog: So this is interesting. I don't think it was @slifty's intent that these be copied into the _project_ `.gitconfig`. These are powerful commands that should, IMO, be left up to each developer to add to their own personal `~/.gitconfig`. I'd recommend removing this from the repo.

reefdog (changes requested):
I can't really review this (aside from the same .gitconfig comment as @oneroyalace) atm. ~Running tests bails out (w/27 errors) because of `Error: 0 returned from external Zorki server`~ (**Edit:** whoops, I didn't have Hypatia running locally despite configuring Zeno to use it, mea culpa! Fixing that, the tests are clean. 🎉), and it looks like we're back to having all those old commits gunking up commit history so I can't really tell what changed.

@cguess do you think you could re-apply the relevant commits to a clean branch of master so we can see what's actually part of this PR?

PR #181, opened 2022-04-15
Equalize force nonforce rebase

cguess:
2nd try on equalizing force between Hypatia and zenodotus.

To test:
- Make sure you have the sister branch to this running on Hypatia
- `rails t`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,3 +58,4 @@ coverage
 dump.rdb
 
 .env
+.gitconfig
--- a/.gitignore
+++ b/.gitignore
@@ -58,4 +58,3 @@ coverage
 dump.rdb
 
 .env
-.gitconfig
--- a/app/models/scrape.rb
+++ b/app/models/scrape.rb
@@ -9,7 +9,6 @@ class Scrape < ApplicationRecord
     response = JSON.parse(response)
     archive_item = case self.scrape_type
                    when "instagram"
-                     # debugger
                      Sources::InstagramPost.create_from_zorki_hash(response)
                    when "facebook"
                      Sources::FacebookPost.create_from_forki_hash(response)
--- a/app/models/scrape.rb
+++ b/app/models/scrape.rb
@@ -8,6 +8,7 @@ class Scrape < ApplicationRecord
   def fulfill(response)
     archive_item = case self.scrape_type
                    when "instagram"
+                     # debugger
                      Sources::InstagramPost.create_from_zorki_hash([response])
                    when "facebook"
                      Sources::FacebookPost.create_from_forki_hash([response])
--- a/app/models/scrape.rb
+++ b/app/models/scrape.rb
@@ -6,12 +6,13 @@ class Scrape < ApplicationRecord
   has_one :archive_item, dependent: :destroy
 
   def fulfill(response)
+    response = JSON.parse(response)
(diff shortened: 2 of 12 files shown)


test/models/instagram_user_test.rb
```diff
@@ -1,10 +1,12 @@
...
 
 class InstagramUserTest < ActiveSupport::TestCase
+  @@zorki_user = nil
+
   def setup
-    @zorki_user = InstagramMediaSource.extract(
+    @@zorki_user = InstagramMediaSource.extract(
       "https://www.instagram.com/p/CQDeYPhMJLG/", true
-    ).first["post"]["user"]
+    )["scrape_result"].first["post"]["user"] if @@zorki_user.nil?
```
oneroyalace, line 9: Ah, does this prevent `InstagramMediaSource.extract` from being run before each individual test?

cguess: Yes, since it's the same exact request, this means it only has to run once per total run.

cguess: I've added this pattern to a few tests.

app/models/scrape.rb
```diff
@@ -6,11 +6,13 @@ class Scrape < ApplicationRecord
   has_one :archive_item, dependent: :destroy
 
   def fulfill(response)
+    response = JSON.parse(response)
     archive_item = case self.scrape_type
                    when "instagram"
-                     Sources::InstagramPost.create_from_zorki_hash([response])
+                     # debugger
```
reefdog, line 12: Could you nix all these `#debugger` lines? May as well do it now than just have to remove them later.

.gitignore
```diff
@@ -58,3 +58,4 @@ coverage
 dump.rdb
 
 .env
+.gitconfig
```
reefdog, line 61: Sorry that we're beating this one to death, but: I don't think we need to blanket ignore `.gitconfig` as a rule! I just think that particular command belongs in the user's own config space (for me, `~/.gitconfig`) rather than in the repo's config space. (Side note: I've actually never seen a repo `.gitconfig` so didn't know it was a thing.)

My earlier review was suggesting that you move the `[alias]` block out of the repo's `.gitconfig` into your own userspace git config, not "leave it here but ignore it".

Any objection to that? My hesitance about this approach is I don't know if `.gitconfig` in the repo has a valid use case we may want in the future, and will have forgotten why we ignored it in the first place.

reefdog (changes requested):
@cguess This runs great and the tests seem to pass. 🎉

```
Finished in 852.076312s, 0.1056 runs/s, 0.2817 assertions/s.
90 runs, 240 assertions, 0 failures, 0 errors, 0 skips
Coverage report generated for Minitest to /Users/justin/Projects/duke/zenodotus/coverage. 878 / 1011 LOC (86.84%) covered.
```

I do have two small requests, but won't need a re-review after them.

oneroyalace (approved):
LGTM

.gitignore
cguess: Point taken and agreed. I'll fix it.

app/models/scrape.rb
cguess: oh damn it!

PR #184, opened 2022-04-21
Condense environment variable list

oneroyalace:
This PR removes the `ZORKI_AUTH_KEY`/`FORKI_AUTH_KEY` and `ZORKI_SERVER_URL`/`FORKI_SERVER_URL` environment variables and replace them with `HYPATIA_AUTH_KEY`/`HYPATIA_SERVER_URL`. 

It also updates `config/intializers/figaro.rb` to require the presence of all of Zenodotus' environment variables. 

Changes are documented in the developer README.

# Testing
- `rails t`

_Note: before testing, make sure you've set all environment variables. Before merging this PR, we may need to create a shared Tech&Check YouTube API key._

### DIFF
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -2,6 +2,20 @@
 
 ## Environment Setup
 
+### Environment variables
+To run Zenodotus, you'll need to set the following environment variables. Ask a dev on the team to provide you access to them. 
+- `INSTAGRAM_USER_NAME`
+- `INSTAGRAM_PASSWORD`
+- `FACEBOOK_EMAIL`
+- `FACEBOOK_PASSWORD`
+- `TWITTER_BEARER_TOKEN`
+- `YOUTUBE_API_KEY`
+- `HYPATIA_SERVER_URL`
+- `HYPATIA_AUTH_KEY`
+- `secret_key_base` (Devise secret)
+- `KEY_ENCRYPTION_SALT`
+- `URL`
+
 ### Linters
 
 We use two linters to insure that all code looks and acts the same. These enforeced styles may be different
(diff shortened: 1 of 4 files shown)


docs/DEVELOPMENT.md
```diff
@@ -2,6 +2,20 @@
 
 ## Environment Setup
 
+### Environment variables
+To run Zenodotus, you'll need to set the following environment variables. Ask a dev on the team to provide you access to them. 
+- `INSTAGRAM_USER_NAME`
+- `INSTAGRAM_PASSWORD`
+- `FACEBOOK_EMAIL`
+- `FACEBOOK_PASSWORD`
```
reefdog, line 10: Do we actually need the IG and FB credentials here in Zenodotus? I thought the task of scraping those were shunted over to Hypatia?

docs/DEVELOPMENT.md
```diff
@@ -2,6 +2,20 @@
...
+- `TWITTER_BEARER_TOKEN`
+- `YOUTUBE_API_KEY`
```
reefdog, line 12: Same question as above re: whether Zeno needs the YouTube key or just Hypatia.

reefdog (commented):
This seems to work great, and I was able to fire up the app and run it and such once I'd added the new required vars.

However, I did leave a few inline questions about whether we actually want/need to require the IG, FB, and YT credentials here in Zeno. Isn't the task of scraping those shunted off to Hypatia? Why do we need those configged here?

docs/DEVELOPMENT.md
oneroyalace: Thaaaat's a great point.

oneroyalace: No we do not need them

docs/DEVELOPMENT.md
oneroyalace: You are once again correct :)

reefdog (approved):
Huzzah! This looks great, thanks for removing those unused keys.

Issue #188, opened 2022-04-21
Reduce/eliminate our Node use


reefdog:
After some Slack chatter with @cguess, it's… possible that we don't actually need Node/yarn on this project at all.

- Rails 7 defaults to a Webpack-free, Node-less system using [importmap](https://github.com/rails/importmap-rails/).
- The libraries we're using that used to be delivered via npm packages (Turbo, Stimulus, Tailwind) are provided via gems now.

This issue is for discussing (if necessary) any consequences, tracking any complications, and then actually doing it.

Among the unknowns:

- How much will we regret only having access to the npm ecosystem through the importmap pinning? E.g., we can no longer [patch packages](https://github.com/ds300/patch-package) to resolve issues like [this one with Flowbite+Turbo](https://github.com/themesberg/flowbite/issues/88) (unless we do something _real_ old-school like just include the package in a `vendor/` dir); but maybe those are dangerous anti-patterns anyway?
- ESLint: Maybe this is the single thing we keep Node around for? Reducing our surface area to just dev dependencies would still be a win.

reefdog:
Once we've done this, I think we can also remove the `cssbundling-rails` gem. [See this FAQ note.](https://github.com/rails/cssbundling-rails/blob/main/README.md#how-does-this-compare-to-tailwindcss-rails-and-dartsass-rails)

reefdog:
I'm not going to rush this because there are more important things to do (although this _is_ important to tackle soon so that our codebase has a very clear MO), but I did just do a quick test of the following changes, and it all worked just fine:

- **Primary:** Delete `package.json`, `yarn.lock`, and `node_modules/`
- **Cleanup:**
   - Remove the Node reference from `.tool-versions`
   - Delete `babel.config.js`, `postcss.config.js`, and `tsconfig.json`
- **Unknown:**
   - Delete Node details from Heroku's `app.json`
   - Delete Node details from Docker's `docker-compose.yml` and `Dockerfile` (not that we're using it)

reefdog:
Well, apologies for the whiplash, but I no longer thing we should do this, and in fact think we should revert back entirely to a bundled setup.

**The pragmatic justification** is that I think we should use Flowbite, and we simply can't use Flowbite correctly with importmaps. Full stop. [DHH says as much here.](https://github.com/rails/tailwindcss-rails/issues/159#issuecomment-1082881077)
[...]

PR #189, opened 2022-04-22
Remove Tailwind’s Node package and related pieces

### LINKED ISSUE #187: Remove Tailwind npm package and all its detritus
We may have started out using the Tailwind CSS npm package, but we now use the [Tailwind CSS gem](https://github.com/rails/tailwindcss-rails). Consequently, we can and should remove it and any related files.

reefdog:
This PR removes the Tailwind Node package and any configuration or build stuff related to it. It wasn't being intentionally used anyway; we now use the Tailwind CSS gem and are moving toward a Webpacker-less future (#188).

I wish this PR could be chopped up into more small commits rather than one big one, but subscribing to a loose theory of individual commits not being knowingly broken required doing it all as one commit. The individual steps taken were:

1. The Tailwind CSS npm package was removed from `package.json`, and the lockfile updated accordingly.
2. The vestigial config files specific to that package (i.e., unused by the Tailwind CSS gem) were removed, along with an unused SCSS file that references more Tailwind.
3. With the npm package gone, the Yarn `build:css` task will fail (unless I do something like fill it with a noop), so that’s removed.
4. With the Yarn scripts gone, the `cssbundling-rails` gem complains during asset compilation, so that gem is commented out and the lockfile updated accordingly. (We don't want it anyway.)

It also includes a change to the README that instructs new devs to precompile assets, to get around some weirdness explained in the description of 9491976a61e837e99900f7d85e6c791f4b4754d3.

### Testing instructions:

1. `yarn install`
2. `bundle install`
3. Delete your `{repo}/public/assets` and  `{repo}/assets/builds` directories
4. `rails assets:precompile` ← this should actually be a first-run development step at all times, will add to README
6. `./bin/dev`
7. Check out the site, click around. If it's not a CSS-less void, all is well!

Closes #187


oneroyalace (approved):
LGTM

cguess (approved):
WORKS!!!!

PR #190, opened 2022-04-22
Implement video hashing

cguess:
This add hashing of video previews and restructures where hashes live to make it easier to index later on.

On the front end you won't notice anything different, as it's all reflected in data structures.

To test:
bundle install
rails db:migrate
rails t


oneroyalace (approved):
LGTM

PR #192, opened 2022-04-26
Revert back to CSS/JS bundling from importmaps

reefdog:
_The background to this issue has been discussed ad nauseum in issues like #188 and on Slack, so I won't go into the full justification here._

This PR removes our use of `importmap-rails` and goes back to the boring old days of asset-bundling via a JS bundler, though in this case [esbuild](https://esbuild.github.io/) rather than Webpack because it's light and fast.

The commits are organized pretty atomically, so I recommend poking through them individually if you're curious about individual changes. fe06b3779f9c9f10143baf022fb9e96e70707be4 was huge by necessity, but I put a lot of clarifying detail into its commit message.

**Testing:**
- Remove `node_modules/`
- Empty `app/assets/builds/` and `public/assets/` to make sure old assets aren't hanging around
- `yarn install`
- `bundle install`
- Run and test the app using one of these two methods:
   - `./bin/dev` will fire up the CSS/JS bundler alongside Puma, so no more work needed; **this is preferred for testing**
   - `rails s` will need `rails assets:precompile` to be run first; then all will be well, though CSS/JS changes won't compiled automatically (I don't think)


cguess:
This breaks when using JS it seems. To test: go to home page (after logging in), click the "+" button. It should put a modal popup over the content, instead it redirects to another page. Can you make sure the Stimulus code is being compiled?

cguess (changes requested):
See previous comment about JS not working.

reefdog:
@cguess Hrm, it's definitely working for me, and verified that the modal is loaded via an XHR `fetch` and not a `GET`.

Can you try emptying `/app/assets/build` and `/public/assets` and trying again? Also, are you booting with `./bin/dev` or `rails s`?

PR #201, opened 2022-04-29
Add video search

cguess:
Add video search

This adds full reverse video search with the same interface as we
already had for images. From the user's standpoint it's the same, but
the backend is all sorts of new things.

This utilizes a new gem I created: [zelkova](https://github.com/cguess/zelkova) Which builds the search tree. Feel free to include the gem in this review if you'd like, or just for reference.

To test:
rails db:migrate
rails db:reset
rails t

start the server, click "media search" and then upload a video to see if
anything shows up. You may consider archiving a youtube video, and then
searching using the same video file to make sure it returns stuff.


Gemfile
```diff
@@ -164,3 +164,6 @@ gem "dotenv-rails"
 
 # Code coverage so we can check if our tests actually cover everything
 gem "simplecov", require: false
+
+# This enables reverse image and video searching
+gem "zelkova", path: "~/Repositories/zelkova"
```
oneroyalace, line 169: Local path!

app/models/image_search.rb
```diff
@@ -15,17 +42,32 @@ class ImageSearch < ApplicationRecord
...
+      end
+
+      videos = video_hashes.map do |video_hash|
+        hash = ImageHash.find(video_hash[:node].metadata[:id])
+        { video: hash.archive_item, score: video_hash[:distance] }
+      end
 
-    images.sort do |image1, image2|
-      image1[:score] <=> image2[:score]
+      videos.sort_by! { |video| video[:score] }
```
oneroyalace, line 69: Should we note here that low score = low distance = better match? I usually assume that high score = better match. 

https://github.com/cguess/zelkova/blob/631cfbff1f150606cc75dba53b9040f62da70250/lib/zelkova/graph.rb#L87-L88

oneroyalace (changes requested):

config/initializers/zelkova.rb
```diff
@@ -0,0 +1,12 @@
+# typed: strict
+require "zelkova"
+
+Rails.configuration.after_initialize do
+  # Build up the graph
+  graph = Zelkova.graph
```
oneroyalace, line 6: Hm, I error out on lines 6 and 7. Should they just be:
```
graph = Zelkova::Graph.new
```

I don't see a radius attribute in `Zelkova::Graph`
https://github.com/cguess/zelkova/blob/631cfbff1f150606cc75dba53b9040f62da70250/lib/zelkova/graph.rb

oneroyalace (changes requested):

app/models/image_search.rb
reefdog: Oof, if that's the case, and if it's possible, I think we should rename `score` to `distance` or something. That would beat having to document usage, since it's possible that will repeatedly confuse people.

app/models/image_search.rb
```diff
@@ -2,30 +2,73 @@
...
+      images
+    else
+      video_hashes = []
+      self.dhashes.each do |dhash|
+        video_hashes = video_hashes | Zelkova.graph.search(dhash["dhash"])
+      end
+
+      videos = video_hashes.map do |video_hash|
+        hash = ImageHash.find(video_hash[:node].metadata[:id])
+        { video: hash.archive_item, score: video_hash[:distance] }
```
reefdog, line 67: If renaming `score` to `distance` in the returned object is too high-impact, I wonder if we could do something else here, like invert the distance to generate an actual score. (Like, if distance is `0-1`, we could do like `score = 1-distance` or something? Basically I agree with @oneroyalace's concern about the expected polarity of `score`.)

cguess: let's change it

oneroyalace (approved):
LGTM

app/models/image_search.rb
```diff
@@ -2,30 +2,73 @@
...
+  sig { params(media_item: ActionDispatch::Http::UploadedFile, current_user: User).returns(ImageSearch) }
+  def self.create_with_media_item(media_item, current_user)
+    mime = IO.popen(
+      ["file", "--brief", "--mime-type", media_item.path],
+      in: :close, err: :close
+    ) { |io| io.read.chomp }
+
+    search =  case mime.split("/").first
+              when "video"
+                ImageSearch.create!({ video: media_item, user: current_user })
```
reefdog, line 26: If we haven't already, let's open a (low-priority) issue to rename the `ImageSearch` model to `MediaSearch`, for reasons that should be obvious lol.

app/models/image_search.rb
```diff
@@ -2,30 +2,73 @@
...
-      next if archive_item.images.empty?
-      image_hash = archive_item.image_hashes.first
+    # For images, we do our thing
+    if self.image.nil? == false
+      image_hashes = Zelkova.graph.search(self.dhashes.first)
 
-      dhash_score = Eikon::Comparator.compare(self.dhash, image_hash.dhash)
-      { image: image, score: dhash_score }
-    end
+      # For videos we have to loop
```
reefdog, line 50: Why is this comment up here? Feels like it should be down with the actual video logic?

app/models/image_search.rb
```diff
@@ -2,30 +2,73 @@
...
+
+
+      images = image_hashes.map do |image_hash|
+        hash = ImageHash.find(image_hash[:node].metadata[:id])
+        { image: hash.archive_item, distance: image_hash[:distance] }
+      end
+
+      images
+    else
```
reefdog, line 59: I do wonder, since we have `self.videos`, if rather than this video-searching being just the `else` to the image-searching, it would make more sense if it were more verbose:

```ruby
if self.image.nil? == false
  # Image stuff
end

if self.video.nil? == false
  # Video stuff
end
```

There's no *functional* improvement there, but it feels a little weird that we're mode-switching into video search just based on the absence of an image search. (I think making it an `elsif` would also work, though I kinda like this explicitness.)

reefdog (changes requested):
I have just one code request (noted inline), but I should also note that testing in-browser (by locally downloading the Tyler the Creator video from Twitter, then uploading that back as the search) returned two search results, both for the correct archive item.

I was also unable to archive a new YouTube video to use for testing, but I'll assume that's unrelated to this and will figure it out separately.

reefdog:
@cguess if you can merge the latest upstream in here, I can finish testing it out!

reefdog:
I'm running into the following two reproducible issues:

**1. Duplicate search results**
1. `rails db:reset` (for a clean database)
2. `youtube-dl https://t.co/It4oIhHk2C` to download the Lumberjack clip from Twitter
3. Perform a media search with the resulting download

What should happen: you get a search result!
What does happen: you get two search results:
duplicate-results

**2. YouTube search fails on missing template**
1. Add a YouTube video to the archive ([example](https://www.youtube.com/watch?v=Df7UtQTFUMQ))
2. `youtube-dl -f 22 https://www.youtube.com/watch?v=Df7UtQTFUMQ` to download it (in MP4 format, since the default is MKV which we don't support)
3. Perform a media search with the resulting download

What should happen: you get a search result!
What does happen: you get a Rails error about the missing YouTube post template:
missing-post-template

oneroyalace:
Linking this here: https://github.com/TechAndCheck/zenodotus/issues/213

app/models/concerns/dhashable.rb
```diff
@@ -52,6 +53,23 @@ def generate_dhashes_for_attached_media
...
+    media_item = self.image
+    media_item = self.video if media_item.nil?
+
+    media_item.open
+    tempfile_path = media_item.tempfile.path
+
+    if self.video.nil?
+      self.dhashes = [Eikon.dhash_for_image(tempfile_path)]
+    else
+      self.dhashes = Eikon.dhash_for_video(tempfile_path)
```
reefdog, line 69: So this may be self-evident to someone who understands what's going in here, but why this is just a method call, while the image one a few lines up is forced into an array, is the sort of thing it'd be nice to clarify in a comment.

reefdog (approved):
This looks and acts mergeable! I do have a few inline comments about code clarity concerns, but I'll let you prioritize those. 👍

app/models/concerns/dhashable.rb
cguess: it's because image returns just a single image and video returns an array. A comment would be good, you're right I'll add now

cguess: done

app/views/image_search/_search_result_list_item_tweet.html.erb
```diff
@@ -1,6 +1,9 @@
+<% result = result[:image] if result.has_key?(:image) %>
+<% result = result[:video] if result.has_key?(:video) %>
```
reefdog, line 2: 🚨 This line will trigger an error if the prior line succeeded, because `result` has been changed and no longer responds to the `has_key?` method.

We probably shouldn't reassign `result`.

reefdog: Let's go with something like this:

```erb
<% search_result = result[:image] if result.has_key?(:image) %>
<% search_result = result[:video] if result.has_key?(:video) %>
```

(Plus change the references elsewhere in the template of course.)

PR #207, opened 2022-05-03
Add S3 as backing storage

cguess:
This adds the ability to use S3 as the backing storage for all of our
archived content. This is because Zeno runs on ephemeral containers and
so every restart would erase everything we have.

This adds the aws gem to support everything

This also adds three ENV variable. `USE_S3_DEV_TEST` which if set to
"true" then requires `AWS_ACCESS_KEY` and `AWS_ACCESS_SECRET` for dev
and test to test the S3 backing. Production will always require it.

To test:
- bundle install
- rails t

to test s3:
- bundle install
- Set `USE_S3_DEV_TEST` to `"true"` and set `AWS_ACCESS_KEY` and
  `AWS_ACCESS_SECRET` (in 1Password)
- rails t

Affect #207 (but doesn't finish it)


config/application.rb
```diff
@@ -4,7 +4,6 @@
 
 # Require the gems listed in Gemfile, including any gems
 # you've limited to :test, :development, or :production.
-puts Rails.groups
```
reefdog, line 7: lol oh _that's_ why it always outputs those when I boot the app!

config/initializers/shrine.rb
```diff
@@ -1,28 +1,49 @@
 # typed: ignore
 require "shrine"
 
+require "shrine/storage/s3"
```
reefdog, line 4: Possibly a silly question but is this necessary in addition to the `shrine` already required above it?

config/application.rb
cguess: yea, sorry, it was vestigial from a thing way back.

config/initializers/shrine.rb
cguess: It is, Shrine's designed so you don't load a bunch of extra stuff

config/initializers/figaro.rb
```diff
@@ -10,3 +10,8 @@
 
 # The URL of the currently running server for Hypatia callbacks
 Figaro.require_keys("URL")
+
+if Figaro.env.USE_S3_DEV_TEST == "true" || Rails.env == "production"
```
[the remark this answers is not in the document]
cguess: this was a mistake, we should use strings for env variables. Fixing

config/initializers/shrine.rb
```diff
@@ -1,28 +1,49 @@
...
   }
 when "production"
   require "shrine/storage/file_system"
 
   Shrine.storages = {
     cache: Shrine::Storage::FileSystem.new("public", prefix: "uploads/cache"), # temporary
-    store: Shrine::Storage::FileSystem.new("public", prefix: "uploads"),       # permanent
+    store: s3_permanent,                                                       # permanent
   }
 end
```
reefdog, line 49: @cguess With the addition of all the conditions and threads here, I had to sort of map it out for myself and ended up refactoring this a bit to use three utility methods. This seems a bit clearer to me, what do you think? (Dunno if we need the explicit `return`s or not.)

```ruby
# typed: ignore
require "shrine"
require "shrine/storage/file_system"
require "shrine/storage/memory"
require "shrine/storage/s3"

# Available bucket names are `zenodotus-testing` (24-hour expiration) and `zenodotus-production` (permanant)
def make_s3_bucket(bucket_name)
  Shrine::Storage::S3.new(
    bucket: bucket_name, # required
    region: "us-east-1", # required
    access_key_id: Figaro.env.AWS_ACCESS_KEY,
    secret_access_key: Figaro.env.AWS_ACCESS_SECRET,
  )
end

def make_shrine_cache
  case Rails.env
  when "development"
    return Shrine::Storage::FileSystem.new("public", prefix: "uploads/cache")
  when "test"
    return Shrine::Storage::Memory.new
  when "production"
    return Shrine::Storage::FileSystem.new("public", prefix: "uploads/cache")
  # Do we want to specify a fallback for safety?
end

def make_shrine_storage
  case Rails.env
  when "development"
    return Figaro.env.USE_S3_DEV_TEST ? make_s3_bucket("zenodotus-testing") : Shrine::Storage::FileSystem.new("public", prefix: "uploads")
  when "test"
    return Figaro.env.USE_S3_DEV_TEST ? make_s3_bucket("zenodotus-testing") : Shrine::Storage::Memory.new
  when "production"
    return make_s3_bucket("zenodotus-production")
  # Do we want to specify a fallback for safety?
end

Shrine.storages = {
  cache: make_shrine_cache(),   # temporary
  store: make_shrine_storage(), # permanent
}

Shrine.plugin :activerecord
Shrine.plugin :cached_attachment_data # for retaining the cached file across form redisplays
Shrine.plugin :restore_cached_data # re-extract metadata when attaching a cached file
Shrine.plugin :tempfile
Shrine.plugin :derivatives
```

cguess: looks great! regarding the comment: no, those are the only three that ever come up in a Rails app

reefdog: Cool. I'll test both before and after this change to make sure there are no regressions, then is it cool if I commit to this branch?

reefdog: Also do you prefer those explicit `return` statements in each case, or that I ditch them?

oneroyalace: Oh neat. I think someone mentioned that shrine had built-in s3 support yesterday, but I glossed over that. Awesome

reefdog (approved):
This seems to be working for me, btw! I did realize we don't actually have a test that ensures the env var is being respected. So really what I can say is: this works whether that var is set to true or not. If it's worthwhile to add a test for it, let's do it. Otherwise, go forth.

PR #208, opened 2022-05-04
Fix youtube scraper

oneroyalace:
This PR resolves a bug that prevented users from archiving YouTube videos via the Web UI. 

 The bug was was caused by a missing switch case in the `Scrape::fulfill` method, and we've resolved it by removing switch/case logic and implementing some metaprogramming logic. 

Bug bounty goes to @reefdog, hot sauce pourer extraordinaire. 

# Testing
- `bundle install`
- `rails t`


cguess:
Note: This requires a `bundle install` to bump up Eikon

PR #214, opened 2022-05-09
210 media source generator

oneroyalace:
This PR updates the `MediaSource` rails generator to bring it into line with new type signatures and methods. It also updates comments and organization in the individual `MediaSource` class files. 

## Callouts

I've changed the type signature of the `valid_host_name` method so that it no longer accepts lone Strings or `nil` objects. The return type of `valid_host_name` must now be `Array[String]`. Thoughts? All scrapers pass arrays right now. If we don't want to validate hosts in a future scraper, we can leave the array empty. 

Someday we should consider moving Birdsong over to Hypatia. The move will slow down scraping, but will bring `TwitterMediaSource` into line with the rest of the `MediaSource` subclasses. 

## Testing
- `rails t test/media_sources/` should test all changes here, but feel free to run the whole suite.  
- To test the generators, try running `rails generate media_source Mastodon` from the command line. Check out the newly generated files. Do they look okay?


app/media_sources/facebook_media_source.rb
```diff
@@ -94,12 +83,26 @@ def retrieve_facebook_post!
...
+
+    # Validate that a url links to a Facebook post
+    #
+    # @!scope class
+    # @!visibility private
+    # @params url [String] a url to check if it's a valid Facebook post url
+    # @return [Boolean] if the string validates or not
+    sig { params(url: String).returns(T::Boolean) }
+    def self.validate_facebook_post_url(url)
+      return true if /facebook.com\//.match?(url)
```
reefdog, line 102: This seems a little bit too loose to me, as it would match falsified domains like `totallynotfacebook.com`. I think we do want to be flexible enough to catch both HTTP and HTTPS, and any subdomains (incl naked domain), but let's tighten it up more than this!

reefdog, line 102: This seems a little bit too loose to me, as it would match falsified domains like `totallynotfacebook.com`. I think we do want to be flexible enough to catch both HTTP and HTTPS, and any subdomains (incl naked domain), but let's tighten it up more than this!

reefdog: Sigh, nothing is as easy as I'd hope, lol:

```shell
irb(main):004:0> ['facebook.com', 'http://facebook.com', 'https://facebook.com', 'https://www.facebook.com', 'http://www.facebook.com', 'foo.facebook.com'].map { |domain| puts
 "#{domain}: #{URI(domain).host == 'facebook.com'}"  }
facebook.com: false
http://facebook.com: true                                     
https://facebook.com: true                                    
https://www.facebook.com: false                               
http://www.facebook.com: false                                
foo.facebook.com: false                                       
```

reefdog: Sigh, nothing is as easy as I'd hope, lol:

```shell
irb(main):004:0> ['facebook.com', 'http://facebook.com', 'https://facebook.com', 'https://www.facebook.com', 'http://www.facebook.com', 'foo.facebook.com'].map { |domain| puts
 "#{domain}: #{URI(domain).host == 'facebook.com'}"  }
facebook.com: false
http://facebook.com: true                                     
https://facebook.com: true                                    
https://www.facebook.com: false                               
http://www.facebook.com: false                                
foo.facebook.com: false                                       
```

oneroyalace:
May come back to this later

PR #215, opened 2022-05-19
Clean up MediaReview ingestion

cguess:
This cleans up MediaReview ingestion automatically from FactStream.
We're still not brining in every element of mediareview, but it's a lot.
THis will do for now as a proof of concept on the auto archiving.

To test:
- rails db:migrate
- make sure you have Hypatia running and https://github.com/TechAndCheck/hypatia/pull/29 is either checked out or merged to master
- From here, honestly talk to chris, it involves the factstream staging
  server and gets extremely complex

### DIFF
--- a/app/models/scrape.rb
+++ b/app/models/scrape.rb
@@ -11,7 +11,7 @@ class Scrape < ApplicationRecord
     archive_item = ArchiveItem.model_for_url(self.url).create_from_hash(response)
 
     # Update the previously created MediaReview item with the now archived stuff
-    media_review_item = MediaReview.find_by(original_media_link: self.url)
+    media_review_item = MediaReview.find_by(original_media_link: self.url, archive_item_id: nil)
     media_review_item.update({ archive_item_id: archive_item.first.id }) unless media_review_item.nil?
 
     self.update!({ fulfilled: true, archive_item: archive_item.first })
--- a/test/controllers/ingest_controller_test.rb
+++ b/test/controllers/ingest_controller_test.rb
@@ -103,7 +103,8 @@ class IngestControllerTest < ActionDispatch::IntegrationTest
     assert_equal "Successfully archived media object", json["response"]
     assert_not_empty json["media_object_id"]
 
-    post = ArchiveItem.find(json["media_object_id"])
+    # We'll make sure the media review is correctly saved here since archive is async
+    post = MediaReview.find(json["media_object_id"])
     assert_not_nil post
   end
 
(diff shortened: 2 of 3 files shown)


cguess:
Some steps on testing:
1. Spin up Hypatia (after all the PRs are merged)
2. Spin up Sidekiq in the Hypatia directory to handle its jobs
3. Pull this branch of zenodotus down
4. rails db:migrate
5. Spring up Sidekiq in this branch of Zenodotus
6. Run `rails t` to make sure it's all working
7. Start up page kite or some other url redirector and make sure that the zenodotus instance is set to accept  host names from your choosing
8. Confirm you can access the running zenodotus instance with the url you've run
9. Log into Heroku and go to the FactStream project
10. Open up the setting for the `factstream-staging` server.
11. Look into the environment variables and set the `ZENODOTUS_SERVER_URL` (or whatever it is, it's obvious) to the URL redirect you're using
12. Log into staging.factstream.co
13. Fill out a media review, making sure to also fill out a claim review for it (we're not supporting solo media review ingestion yet)
14. Everything should kick off on submission running through sending to Zeno, sending to Hypatia, backwards.

Note that in fact stream this isn't a separate job yet, so if there's a bug in sending to Zeno the submission will fail. This will be fixed soon but is tangental to this PR.

app/models/scrape.rb
```diff
@@ -10,6 +10,10 @@ def fulfill(response)
 
     archive_item = ArchiveItem.model_for_url(self.url).create_from_hash(response)
 
+    # Update the previously created MediaReview item with the now archived stuff
+    media_review_item = MediaReview.find_by(original_media_link: self.url)
```
reefdog, line 14: Just to be clear, we have a uniqueness restraint on `MediaReview`'s `original_media_link`?

reefdog: Per some Slack convo with Chris:

- We do have (and want to maintain) a 1:1/`belongs_to` relationship between `MediaReview` and `ArchiveItem`
- We do (and want to continue) redundantly scrape identical URLs, because we can't guarantee the content hasn't changed since the last time we scraped (think: edit posts)
- However, yes, this line is too broad and would mean new instances of the scrape were attached not to the newly-created MR, but to the first MR that happened to exist for that URL
- Ideally, we'd do something like have the MR's ID carried along the entire scrape process and be used for referencing the MR once the scrape was complete
- But that's not practical atm and a just-fine stopgap that reduces the problematic surface area is to **add an `archive_item_id: null` condition** to this query

reefdog (changes requested):
This is so cool. Watching it happen was like magic.

My only requested change is the addition of the `archive_item_id: null` scope on the query that associates completed scrapes with their presumed `MediaReview`, to limit the changes of accidentally attaching them to a previous MR for the same URL, which I believe you're already working on.

Once that's done I should probably give it a final test to make sure it works, but then 🚀!

reefdog (approved):
With the latest changes, this works great and tests clean. 🚀

Issue #205, opened 2022-05-02
Set up third party storage and upload


cguess:
This is an archive, however, we use ephemeral containers right now for everything (scraping and storage). This is a problem, since every time the server resets all the data will disappear (on Zenodotus) and fill up all the bandwidth (on Hypatia). Instead, it'd make sense to use S3 (maybe something else?) as a storage backend. This would mean Hypatia uploading results to S3 and then sending an id in the response.

Pros:
- This has the benefit of dramatically lowering our REST call size (which is unreliable anyways)
- don't lose our data (probably number 1 reason)

Cons:
- More steps
- having to deal with AWS credentialing

cguess:
This graph describes the work flow that I'm seeing (link to edit below the image).
![Untitled-3](https://user-images.githubusercontent.com/322626/166311001-9edf504d-8fb6-4588-ba00-939f8cccb428.png)
https://sequencediagram.org/index.html#initialData=A4QwTgLglgxloDsIAIBaBTBB7AJliArgM4BQoks8ISyAEgJ6jQgk4gQgBGIR6yAggHUAysmEBmOo3ZQQyAEIEYAa3QRW7Ljz5DREtJlz5iCpavUloEADZ9hMMCGDocyALIvZYiFkcBzPgAxaywAdxISDGw8QiJkAFoAPgYmWQAuZAAldABHAnQiFHtHZxIUmRB4pKijWIyAJgAGRuQAeQBpMulmBOTu9LEHJzsoCHQu1Lkk3TFJcp7FFTUMgFVgEJBXIiHnVwAzKFsSbDHkLAA3dDABEVmpSdMliAyAFQALKDjOMzVkN55kHIsJwAFboGAoaAAW3QWAIKGAWGssHoAB1sHtAcg9uhQsg2PQ4kQsMgIP8UKE+HgEAByCDo-6XUkk0K+MD0QGcOEoHDoWzQBB+M6g8EQUgzfTzLyLcy9KUgDIy36tEUQ5AASQAIhMKtVDDFiBlisNsVAEJ83i5kDAQNZrNwVOjQqM3shvk9hWCYMgoPtfKajjUDUQADzTW6S-pyJXPZCasIIDZbHZWmE4WSRfXGIh66LZjIABTAWBgBTiaYzQez4b0kirsUe5iNPjAfArcj2-oAXlmGwAKMmfZDEApiN5gM3oGlfAiHAV3HHsAitogASiAA

[...]

PR #223, opened 2022-05-31
Add S3 downloading for Facebook Posts

cguess:
This adds the ability to receive facebook posts that have been uploaded
to s3 instead of sent via Base64

NOTE: This will eventually include other models as well, but it's best to check it now so that I can copy/paste over to the others.

This adds two new AWS settings
```
AWS_S3_PATH_HYPATIA: ""
AWS_S3_BUCKET_NAME_HYPATIA: "zenodotus-testing"
```

For testing you can just copy this code block to `config/application.yml`

NOTE: This changes the old AWS environment variable names as well, please update
them to the new ones outlines in `/config/initializers/fiagro.rb`

To test:
1. Make sure hypatia is set up with master and all the AWS settings are
   correct
1. Insure you've updated your AWS settings, and added the new keys
1. `rails t test/models/facebook_post_test.rb`

### DIFF
--- a/app/controllers/ingest_controller.rb
+++ b/app/controllers/ingest_controller.rb
@@ -173,14 +173,14 @@ class IngestController < ApplicationController
     }
   end
 
-  private
+private
 
-    # Validate MediaReview that was passed in
-    sig { params(media_review: Hash).returns(T::Boolean) }
-    def validate_media_review(media_review)
-      schema = File.open("public/json-schemas/claim-review-schema.json").read
-      JSONSchemer.schema(schema).valid?(media_review)
-    rescue StandardError
-      false
-    end
+  # Validate MediaReview that was passed in
+  sig { params(media_review: Hash).returns(T::Boolean) }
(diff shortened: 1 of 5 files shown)


reefdog:
It looks like Zenodotus needs `AWS_REGION` set, too. Otherwise, I get seven errors of: 

```
Aws::Errors::MissingRegionError: No region was provided. Configure the `:region` option or export the region name to ENV['AWS_REGION']
```

But with that set, all is well!

reefdog (approved):
This works and looks good! My only note is that I had to add `AWS_REGION` to the `application.yml`, else the tests failed.

Do you intend to configure Figaro to require these new keys when S3 is enabled (or in prod)?

cguess:
> Do you intend to configure Figaro to require these new keys when S3 is enabled (or in prod)?

Yes, I do. Thanks for the note!

reefdog:
@cguess BTW does this PR address #205? It would seem to! If so, may want to tag the description.

PR #229, opened 2022-06-02
Upgrade ESLint and clean up config

### LINKED ISSUE #227: Configure ESLint the way we want
Currently, our [ESLint configuration](https://github.com/TechAndCheck/zenodotus/blob/master/.eslintrc) results in complaints about perfectly valid JS like this:

```javascript
import { Controller } from '@hotwired/stimulus'
export default class extends Controller {
    static targets = [ "foo" ]
}
```

The specific error is:

```
[{
	"resource": "/Users/justin/Projects/duke/zenodotus/app/javascript/controllers/archive_controller.js",
	"owner": "eslint",
	"severity": 8,
	"message": "Parsing error: Unexpected token =",
	"source": "eslint",
	"startLineNumber": 3,
	"startColumn": 20,
	"endLineNumber": 3,
	"endColumn": 20
}]
```

Consequently, we learn to just ignore ESLint, which is Bad.

This issue is for researching what our config should be and applying it to the repo.

reefdog:
Our existing ESLint was out of date and showing parsing errors when using out-of-the-box Stimulus. It was also very React-focused, which we aren’t using. This PR:

- Upgrades ESLint and ESLint-related packages we are using
- Removes the packages we aren’t using
- Converts us from the React-focused `airbnb` plugin to the React-free `airbnb-base` plugin
- Removes React/JSX-related rules

**Testing:**
- **Before installing**, run `eslint app/javascript/controllers/image_search_controller.js`
- ❌ Note that you get a parsing error
- `yarn install`
- Lint that same file
- ✅ Note that you get eight errors, which are actual lint errors

Closes #227

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,18 +1,14 @@
 {
-  "extends": "airbnb",
+  "extends": "airbnb-base",
   "rules": {
     "semi": [2, "never"],
     "max-len": ["error", { "code": 150 }],
     "prefer-destructuring": ["error", {"array": false, "object": true}],
-    "react/destructuring-assignment": [0, "always"],
-    "react/prop-types": [0, "never"],
-    "react/jsx-props-no-spreading": [0],
     "no-alert": [0, "never"],
-    "jsx-a11y/label-has-associated-control": [2, {
-      "controlComponents": ["DatePicker"],
-      "depth": 3
-    }],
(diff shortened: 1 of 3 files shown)


oneroyalace (approved):
LGTM. Testing instructions lead to expected outcome!

Issue #234, opened 2022-06-03
Design and style user pages


reefdog:
This issue is for the designing and styling of user profile pages. These currently will be platform-specific users, since we don't have the concept of an "author" or "person" that ties together accounts on different platforms. (E.g., LeBron James on Twitter is a different page than LeBron James on Instagram.)

reefdog:
This has been done for a while, weird.

PR #233, opened 2022-06-03
Lint all our application JavaScript

reefdog:
_This PR ladders off #220 and shouldn't be merged until that one is._

This PR lints all our JS, and tweaks a few configurations discovered as necessary:

- While function params still can't be reassigned, param _properties_ now can be. [This is just fine](https://eslint.org/docs/rules/no-param-reassign#props) and allows for reasonable things like `function foo(el) { el.innerText = 'Hi!' }`.
- JavaScript files auto-generated by Rails are now being ignored by ESLint.
- After some ping-ponging, I've decided that allowing absolute imports (e.g, `import 'controller'`) from our `app/javascript` directory is a Bad Idea likely to lead to developer confusion (what's ours? what's global/vendor?) and possible name collisions. So, that's been removed from our `jsconfig.json`.

**Testing:**
- On `master`, run `eslint app/javascript`
- ❌ Note the 66 problems
- On `230-lint-javascript`, run `eslint app/javascript`
- ✅ Note the clean bill of health

PR #236, opened 2022-06-16
Refactor archive items page

reefdog:
🚨 _This PR ladders off #233. Please review that one first._

This PR represents a spray of improvements implemented during the implementation of #234 that are not, in fact, directly related to that task. So there's not much to _see_ with this one. Just scan the code (and the commits, which are organized) and call out any weirdness.

This allows the next PR, targeting Twitter user pages, to be very clean.

**Testing:**
- `./bin/dev` and poke around the main page. It shouldn't look much different from before.
- Look at the code. Is it sensible?

### DIFF
--- a/app/assets/stylesheets/archive_items.css
+++ b/app/assets/stylesheets/archive_items.css
@@ -1,95 +1,137 @@
-.archive-item {
-	--profile-photo-size: 4rem;
-	--author-area-gap-x: 1rem;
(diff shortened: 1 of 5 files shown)


cguess (approved):
This is a big one, so it's hard to review line by line, but I trust @reefdog intrinsically so, I'm approving.

PR #237, opened 2022-06-16
Redesign user pages

reefdog:
This PR significantly redesigns the user pages.

**Before:**
before

**After:**
after

⚠️ The "After" screenshot shows a masonry-style layout that actually only works (1) in Firefox and (2) after enabling an experimental feature. :( You'll likely actually see a slightly different and less compact layout.

**Testing:**
- Scan the code, how's it look? Sensible?
- `./bin/dev`, go to the Most Recent Archived Items page, and click on the author.

### DIFF
--- a/app/controllers/twitter_users_controller.rb
+++ b/app/controllers/twitter_users_controller.rb
@@ -4,7 +4,10 @@ class TwitterUsersController < ApplicationController
   sig { void }
   def show
     @twitter_user = Sources::TwitterUser.find(params[:id])
-    @archive_items = Sources::Tweet.where(author_id: @twitter_user.id).includes(%i[images videos])
+    tweet_archive_items = ArchiveItem.includes(archivable_item: [:author, :images, :videos])
+                                     .where(archivable_item_type: "Sources::Tweet")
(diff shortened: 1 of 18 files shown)


app/controllers/twitter_users_controller.rb
```diff
@@ -4,7 +4,10 @@ class TwitterUsersController < ApplicationController
   sig { void }
   def show
     @twitter_user = Sources::TwitterUser.find(params[:id])
-    @archive_items = Sources::Tweet.where(author_id: @twitter_user.id).includes(%i[images videos])
+    tweet_archive_items = ArchiveItem.includes(archivable_item: [:author, :images, :videos])
+                                     .where(archivable_item_type: "Sources::Tweet")
+    tweets_by_author = tweet_archive_items.select { |t| t.tweet.author == @twitter_user }
```
reefdog, line 9: This `select` exists only because I don't know how to construct the deep query as part of the ArchiveItem query conditions. If one of you does, please suggest!

cguess: You can add separate pure SQL commands, but off the top of my head I'm not sure.

cguess (approved):

PR #246, opened 2022-06-22
Fix Facebook Users duplicating

### LINKED ISSUE #239: Single Facebook author imported as multiple users
I just loaded up https://zenodotus-staging.herokuapp.com and noticed that the two most recently-archived pieces of media (at this moment) are from the same Facebook account, yet if you drill in to their user page, they're treated as distinct Facebook users within the vault. It looks like we may sometimes be creating a new user record for every archive item.

Example of three users that should be one:

- https://zenodotus-staging.herokuapp.com/facebook_users/8274ed34-0445-4494-aa86-959d7f978e7e
- https://zenodotus-staging.herokuapp.com/facebook_users/54c4ddaa-e544-4592-86fc-404b83614cc8
- https://zenodotus-staging.herokuapp.com/facebook_users/3541c332-afbc-424b-ae90-f1893040abda

cguess:
Instead of searching for current users by `facebook_id` when checking if
we should create a new user or just add a new piece of media to
a current user we were looking at `id` which, of course, never matched.

This fixes this error and also introduces `rails users:dedup_facebook`
as a rake task to fix this retroactively as well.

To test:
- Run `rails t test/models/facebook_user_test.rb`
- Run `rails users:dedup_facebook`

closes #239

### DIFF
--- a/app/models/sources/facebook_user.rb
+++ b/app/models/sources/facebook_user.rb
@@ -19,7 +19,7 @@ class Sources::FacebookUser < ApplicationRecord
   def self.create_from_forki_hash(forki_users)
     forki_users.map do |forki_user|
       # First check if the user already exists, if so, return that
-      facebook_user = Sources::FacebookUser.find_by(id: forki_user["id"])
+      facebook_user = Sources::FacebookUser.find_by(facebook_id: forki_user["id"])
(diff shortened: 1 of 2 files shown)


oneroyalace (approved):
LGTM. Very good catch. That's a nasty bug

PR #247, opened 2022-06-22
Hide inaccessible user-related routes and pages

reefdog:
This PR is a "step backwards to go forwards" thing. It reduces our Devise routes down to just session login/logout, since nothing else is wired up or working. This way testers (and ourselves) don't stumble upon them.

Affects #153


reefdog:
Never mind this, it breaks tests all over and was the wrong idea, apologies.

Issue #252, opened 2022-06-23
Facebook previews on user page don't play video


cguess:
If you go to a user's page with a Facebook video post on it we only show the preview with no way to play it. Fix this.

reefdog:
@cguess I can't replicate this. When I go to a user's page, of any kind platform (including Facebook), I'm able to play the video.

We are using the native `<video>` tag and its native controls, which are hidden after a moment of inactivity. But you can click the video area to start playing, or hover (on desktop) to reveal controls.

Can you verify that this issue exists, and maybe we can screenshare about it if so?

reefdog:
Talking w/@cguess, this was meant to be about preview images not videos, thus a dup of #258.

PR #294, opened 2022-08-05
Improve video embedding

### LINKED ISSUE #258: Fix preview images of videos not loading
It *seems* as if the preview images of videos aren't being used when loading the MediaVault's home page and search results. This is problem because that means that the whole video is downloaded before showing the preview, which on slightly slower connections leaves a large white box and nothing else for awhile. On mobile this kills data.

This could be released to the previous issue of some images still being sent over Base64, or it could be an inline issue. Unsure, but we need to look into it ASAP.

Note: You won't notice this when testing locally since the video loads so quickly.

reefdog:
This PR disables video preloading, which happens automatically under certain browser/bandwidth conditions. With preloading allowed, then the browser would sometimes skip the poster and just download the full video; it's not clear what these conditions were, but it's possible we hit a timeout on the `poster` download and so triggered the preload-for-metadata process.

Now, we set`preload="none"` on all `<video>` elements to ensure the browser doesn't begin downloading the videos until the user actually plays the video. This is preferable anyway for us since we are usually showing multiple (sometimes very many) videos per page, with no confidence the user is likely to play any of them, but it should also ensure the browser waits around for the poster image and doesn't try to fetch the video for metadata / first frame.

**To test:**

1. On `master`, load up the Media Vault "home page" with your inspector open to the Network tab, filtered to Media. (Also make sure Disable Cache is checked.)
2. ❌ The videos are transferred even before you play them. (**Note:** You may not _always_ see this happen. The `preload` default is `auto`, which lets the browser decide based on network conditions and magic. The point is, though, it _can_ happen, and we really don't mean for it ever to.)
3. Switch to `258-fix-video-preloading` and perform the same check.
4. ✅ The videos aren't transferred until you play them.

Closes #258

### DIFF
--- a/app/views/archive/_archive_item_media_inline_full.html.erb
+++ b/app/views/archive/_archive_item_media_inline_full.html.erb
@@ -8,6 +8,7 @@
             width="100%"
             height="auto"
             poster="<%= media.video_derivatives[:preview].url %>"
+            preload="none"
             controls
         >
             <source src="<%= media.video_url %>" type="video/mp4">


cguess:
> pending some questions
@reefdog Which questions would you like to discuss? (I was the one who flagged this issue initially)

reefdog:
@cguess Check the two issues (#252 and #258).

cguess (approved):
Works like a damn charm.

One side question: do we want to mute videos on playback by default?

PR #313, opened 2022-08-22
Add Request Access page

### LINKED ISSUE #295: Build registration application system
_This task is part of #296 and follows the design of #268._

Prospective users should be able to apply for an account using a sign-up form.

Necessary components:

- A term that's like "application" but not overloaded like "application"
- `Applicant` model and its validations
- Sign-up page (different from the Devise one)
- Thank you / what's next page

reefdog:
This PR adds the Request Access page to the site. It's not full-featured, as it doesn't yet require email confirmation or allow the user to close the loop on creating a user account, but it's the first step.

**Testing:**
- Run the migrations
- Check tests with `rails t test/models/applicant_test.rb`
- Start the app with `./bin/dev`
- Go to the "Request access" link in the navbar
- Fill out the application and submit

**Notes:**
- I used the `required` attribute on fields that are required, which means the browser will stop the user from submitting a form without them. You can flip them to `required: false` to see Rails validation in action.
- The "Thanks" page is unstyled. That's because it's going to quickly be replaced with the "You must confirm your email" page anyway.
- The language is still being worked on via the Google content doc, thus the placeholders like `[X DURATION]`.

request-access

Closes #295

### DIFF
--- a/app/controllers/applicants_controller.rb
+++ b/app/controllers/applicants_controller.rb
@@ -1,0 +1,48 @@
+class ApplicantsController < ApplicationController
+  sig { void }
+  def new
+    @applicant ||= Applicant.new
+  end
+
+  # A class representing the allowed params into the `create` endpoint
+  class CreateApplicantParams < T::Struct
+    const :name, String
+    const :email, String
+    const :affiliation, T.nilable(String)
+    const :primary_role, T.nilable(String)
+    const :use_case, T.nilable(String)
+    const :accepted_terms, String
+  end
+
+  sig { void }
+  def create
+    # Verify the types of the values provided in the request.
+    # TODO: Pass the output of this to `Applicant.create!()`
+    TypedParams[CreateApplicantParams].new.extract!(applicant_params)
+
+    @applicant = Applicant.new(applicant_params)
+
+    if @applicant.save
+      redirect_to new_applicant_thanks_path
+    else
+      flash.now[:error] = "We weren’t able to save your application. Please be sure to fill out all required fields."
+      render :new, status: :unprocessable_entity
+    end
+end
+
+  sig { void }
+  def thanks; end
+
+private
+
+  def applicant_params
+    params.require(:applicant).permit(
+      :name,
+      :email,
+      :affiliation,
+      :primary_role,
+      :use_case,
+      :accepted_terms
+    )
+  end
+end
--- a/app/controllers/applicants_controller.rb
+++ b/app/controllers/applicants_controller.rb
@@ -25,7 +25,7 @@ class ApplicantsController < ApplicationController
     if @applicant.save
       redirect_to new_applicant_thanks_path
     else
-      flash.now[:error] = "We weren’t able to save your application. Please be sure to fill out all required fields."
+      flash.now[:error] = "We weren’t able to save your application. Please check the form below for errors."
       render :new, status: :unprocessable_entity
     end
 end
--- a/app/models/applicant.rb
+++ b/app/models/applicant.rb
@@ -1,0 +1,36 @@
+class Applicant < ApplicationRecord
+  # A boolean value for terms acceptance stored in the database isn't good enough for auditing,
+  # but a boolean attribute is useful for magic form composition and validation. Thus, we use a
+  # non-database-backed attribute until model creation time, when we generate the timestamp and
+  # version number for database storage.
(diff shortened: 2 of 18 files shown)


app/controllers/applicants_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+    if @applicant.save
+      redirect_to new_applicant_thanks_path
+    else
+      render :new, status: :unprocessable_entity
+    end
+end
+
+  sig { void }
+  def thanks
+    @title_tag = "Thanks"
```
cguess, line 36: Is it necessary to add a variable here or is possible to just hardcode it into the template this renders?

app/models/applicant.rb
```diff
@@ -0,0 +1,33 @@
+class Applicant < ApplicationRecord
+  # A boolean value for terms acceptance stored in the database isn't good enough for auditing,
+  # but a boolean attribute is useful for magic form composition and validation. Thus, we use a
+  # non-database-backed attribute until model creation time, when we generate the timestamp and
+  # version number for database storage.
+  attr_accessor :accepted_terms
+
+  validates :name, presence: true
+  validates :email, format: { with: URI::MailTo::EMAIL_REGEXP }
```
cguess, line 9: Validating emails via regex is usually quite fraught. Just remind me, is this built into Rails or is it a separate thing?

app/models/applicant.rb
```diff
@@ -0,0 +1,33 @@
...
+    self.accepted_terms ||= self.accepted_terms_at.present? && self.accepted_terms_version.present?
+  end
+
+  # Proxy the boolean acceptance of terms to the actual data we need to store on the record.
+  # See the note above on the accessor for why.
+  sig { void }
+  def map_terms_accessor_to_database_values
+    if self.accepted_terms
+      self.accepted_terms_at = Time.now
+      self.accepted_terms_version = TermsOfService::CURRENT_VERSION
```
cguess, line 30: I like this, would have never thought of it myself.

db/migrate/20220808225818_create_applicants.rb
```diff
@@ -0,0 +1,14 @@
+class CreateApplicants < ActiveRecord::Migration[7.0]
+  def change
+    create_table :applicants, id: :uuid do |t|
+      t.string   :name, null: false
+      t.string   :email, null: false
+      t.string   :affiliation
+      t.string   :primary_role
+      t.text     :use_case
+      t.datetime :accepted_terms_at
+      t.string   :accepted_terms_version
```
cguess, line 10: Do we want this to be a string? I think it'd make more sense to make it either a version number or a datetime of when the version went into service. That way, if we need to later, we can more easily compare `>` or `<` to see if we need to get people to agree to new ones.

cguess (approved):
Tentatively approved with a few questions and one code optimization.

Very nice work!

app/models/applicant.rb
reefdog: `URI::MailTo::EMAIL_REGEXP` comes from the Ruby stdlib. I'm happy to pull it if it causes problems, I just do like trying to keep people from shooting themselves in the foot. My understanding is this one follows the email spec _so_ closely that it's actually more broad than reality, e.g. accepting `1234@1234`. So we shouldn't have issues with people's valid addresses being rejected, but it also may be functionally no better than `presence: true`.

Happy to pull it if you'd like!

db/migrate/20220808225818_create_applicants.rb
reefdog: This is a good point… though I'm having trouble thinking about times we'd need something different than just `!=` w/r/t T&Cs. They're either on board with the latest terms or they aren't, and there's no such thing (AFAIK) as having agreed to _future_ terms.

That said, I think "version numbering" the T&Cs as the date they went into service is a fantastic idea, so with your blessing I'll modify this so that it's a `datetime` and so `CURRENT_VERSION` is an actual date.

app/models/applicant.rb
cguess: We can keep it. The way I've always learned to validate email is pretty awful. You send an email and see if it bounces. But that's insanely difficult

reefdog:
@cguess @oneroyalace Changes made (except the email regex), can you give it one last 👀?

cguess (approved):
Good to go!

PR #315, opened 2022-08-24
Remove organizations and related concepts

### LINKED ISSUE #302: Remove concept of organizations from the codebase
At one point, the idea of individuals belonging to organizations was semi-built into the codebase. We no longer intend to build out this functionality, and so should just remove the work to clean up the codebase and reduce semantic collisions as we complete other work. (E.g., `User#admin?` currently determines if a user is an organization admin, but we'd like to use similar terminology to determine if a user is an admin of the site.)

reefdog:
We previously implemented a hierarchy with all users existing under an organization, and organization admins managing the approval/rejection of prospective users, and site super-admins having Power Over All. We are no longer going to use that model; instead, prospective users are Applicants and only super-admins have the power to promote them to Users. Thus, the concepts that can be removed: organizations (`Organization`), organizational admins (`User.admin`), and user approval (`User.approval`).

This PR removes all of that, while updating anything else deprecated by that removal.

- Remove `Organization` model and scaffolding (controller, helper, tests, fixtures, etc.)
- Remove deprecated attributes, e.g. `User.admin`, `User.approved`, and foreign-key references / indices
- Remove deprecated user-approval views, actions, JavaScript, etc.
- Update tests, fixtures, and seeds
- Update authentication methods to treat `super_admin` as `super_user` rather than the organization admin

**Testing:**
- Run `rails db:migrate`
- Run tests
- Log in and poke around as an admin (default seed: `admin@example.com`) and user (`user@example.com`), particularly the settings screens and archiving a URL

Closes #302


db/migrate/20220824134550_remove_organizations.rb
```diff
@@ -0,0 +1,11 @@
+require_relative "20220106154007_create_organizations"
+require_relative "20220107142219_add_organization_to_users"
+require_relative "20210803182641_add_admin_aproved_to_users"
+
+class RemoveOrganizations < ActiveRecord::Migration[7.0]
+  def change
+    revert AddAdminAprovedToUsers
```
oneroyalace, line 7: Whoa, neat. I've never seen this before.

db/seeds.rb
```diff
@@ -8,51 +8,28 @@
...
-  approved: false,
-  admin: false,
-  confirmed_at: Time.now,
-  organization_id: organization.id
+  confirmed_at: Time.now
 },
 # restricted account
 {
-  email: "user1@example.com",
+  email: "restricted@example.com",
```
oneroyalace, line 27: 👍🏿

oneroyalace (commented):
Code LGTM :). Running tests

db/migrate/20220824134550_remove_organizations.rb
reefdog: Me neither! I was about to start writing a big ol manual migration and then thought to check the docs just in case. [Sure enough!](https://guides.rubyonrails.org/active_record_migrations.html#reverting-previous-migrations)

db/seeds.rb
reefdog: This was a very specific line to give a thumbs-up to! 😆

cguess:
When running tests I got the following:
```
# Running:

.........F

Failure:
ApplicantTest#test_can_convert_terms-acceptance_attributes_from_database [/Users/christopher/Repositories/Reporters_Lab/zenodotus/test/models/applicant_test.rb:36]:
Expected false to be truthy.


rails test test/models/applicant_test.rb:35

F

Failure:
ApplicantTest#test_cannot_un-accept_terms [/Users/christopher/Repositories/Reporters_Lab/zenodotus/test/models/applicant_test.rb:49]:
Expected false to be truthy.


rails test test/models/applicant_test.rb:47

E

Error:
ApplicantTest#test_can_edit_applicant_without_triggering_acceptance_errors:
ActiveRecord::RecordInvalid: Validation failed: Accepted terms Terms must be accepted.
    test/models/applicant_test.rb:62:in `block (2 levels) in <class:ApplicantTest>'
    test/models/applicant_test.rb:61:in `block in <class:ApplicantTest>'


rails test test/models/applicant_test.rb:58
```

Is this on me or your code?

cguess (approved):

reefdog:
For history' sake: we realized Chris was running into those errors because I had modified the migrations on the Applicant work in #313 while it was still safely on the branch. Though untangling that did reveal that the migration from _this_ PR isn't reversible, which is… bad! (It runs into NULL constraints when trying to recreate the `organization_id` association apparently.) None of us like having a point of no return beyond which we can't reverse migrations. But! We're also pre-launch so calling it okay.

PR #319, opened 2022-08-25
Remove Twitter to Hypatia

cguess:
This offloads Twitter to Hypatia from here. We're doing this so that all scraping happens away from Zeno and we get a proper separation of concerns.

To test:
- `bundle install`
- `rails t`
- Launch and make sure Hypatia is running and try to archive a Twitter URL. This should kick off to Hypatia instead of running locally. Make sure you have the latest version of Hypatia


test/models/archive_item_test.rb
```diff
@@ -14,9 +14,9 @@ def around
 
   test "destroying a user resets the submitter_id of ArchiveItems it created" do
     sign_in users(:user)
-    Sources::Tweet.create_from_url "https://twitter.com/unsung_son/status/1470963204855578626", users(:user)
+    Sources::Tweet.create_from_url! "https://twitter.com/jack/status/20", users(:user)
     assert_not_nil ArchiveItem.first.submitter
-    User.destroy(users(:user).id)
+    User.destroy(users(:user3).id)
```
reefdog, line 19: I don't think `user3` exists.

cguess: ug.... how the hell did this get in there. it's a merge issue I'm sure.

cguess: fixed, somehow it didn't make it into the force push

app/media_sources/twitter_media_source.rb
```diff
@@ -54,10 +55,54 @@ def initialize(url)
   # @!visibility private
   # @params url [String] a url to grab data for
   # @return [Birdsong::Tweet]
```
reefdog, line 57: Oh, I just noticed that the rdoc (or whatever) `@return` value didn't change here, just the Sorbet sig. It should update too.

app/media_sources/twitter_media_source.rb
```diff
@@ -54,10 +55,54 @@ def initialize(url)
...
+    raise TwitterMediaSource::ExternalServerError if response_body["success"] == false
+
+    true
+  end
+
+  # Scrape the page by sending it to Hypatia and forcing the server to process the job immediately. Should only be used for tests
+  #
+  # @return [Hash]
+  sig { returns(Hash) }
+  def retrieve_tweet!
```
reefdog, line 86: These two methods share so much code, I wonder if the bulk of it could be abstracted out? A little worried they'll deviate in the future when patches land in only one of them.

app/models/sources/tweet.rb
```diff
@@ -74,35 +75,78 @@ def self.create_from_hash(birdsong_tweets, user = nil)
...
-      #   screenshot_attributes = { image: File.open(birdsong_tweet["screenshot_file"], binmode: true) }
+      # image_attributes = birdsong_tweet.image_file_names.map do |image_file_name|
+      #   { image: File.open(image_file_name, binmode: true) }
+      # end
+
+      # video_attributes = birdsong_tweet.video_file_names.map do |video_file_name|
+      #   { video: File.open(video_file_name.first, binmode: true), video_type: birdsong_tweet.video_file_type }
       # end
 
+      # TODO: Uncomment this after Birdsong has been migrated to Hypatia
```
reefdog, line 136: I think we can nix this comment now. :)

reefdog (approved):
This works! 🎉

I have a few inline notes you can choose to ignore.

PR #327, opened 2022-08-30
Make "use case" field required on Request Access form

### LINKED ISSUE #324: Make "use case" a required field on the application form
Use case is vital information, so let's make sure it's required.

### LINKED ISSUE #326: Indicate T&Cs are required with an asterisk
"Accept the terms and conditions" are required, but aren't preemptively indicated as such. Let's add the ol' asterisk there too.

Image

reefdog:
This PR makes the "How do you plan to use the data?" field on the Request Access form required, and moves it to be with the other required fields:

request-access

As shown there, it also adds the "required" asterisk to the "Accept the terms and conditions" checkbox, which was already required functionally.

**Testing:**
- Make sure you can't submit the form without entering a use case (stretch testing: disable `required: false` in the HTML template to trigger the Rails validation)

Closes #324
Closes #326

### DIFF
--- a/app/models/applicant.rb
+++ b/app/models/applicant.rb
@@ -7,6 +7,7 @@ class Applicant < ApplicationRecord
 
   validates :name, presence: true
   validates :email, format: { with: URI::MailTo::EMAIL_REGEXP }
+  validates :use_case, presence: true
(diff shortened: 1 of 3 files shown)


oneroyalace (approved):
LGTM. Tested normally and with `required` disabled

Issue #314, opened 2022-08-23
Stub out Typhoeus requests in tests


oneroyalace:
We currently retrieve live post data from Hypatia during our Zenodotus tests. This takes a long time and makes Zenodotus testing success dependent on Hypatia/scraper/mediasource availability.  Instead of retrieving live post data, we can grab mock post data.

Issue #206, opened 2022-05-03
Mock hypatia for tests


oneroyalace:
Explore the feasibility of mocking Hypatia for Zenodotus tests. 

Requirements: 
1. Keep mocks in sync with Zenodotus/Hypatia APIs (write validators?) 
2. Keep mocked media size in check (select easily compressible media to store in the repo (videos must be > 20s for video search))

PR #330, opened 2022-08-30
Mock hypatia and s3 in tests

oneroyalace:
This PR creates mocks of Hypatia and our `AwsS3Downloader` client and uses those mocks to stub out the methods we use to access Hypatia/S3 in tests. With the intro of this PR, we can run Zenodotus tests without Hypatia/s3 access. 

Instead of scraping media posts to get data for tests, we load that data with mocks. This is, as you might image, much faster. Testing time should drop from 15+ minutes to ~3-4 minutes, most of which will be spent transcoding media. 

Closes https://github.com/TechAndCheck/zenodotus/issues/314, #206 
## 76 files changed?? 
Sorry, there are tons of files changed in this PR. Nearly every test in our suite has been touched up. But here are some descriptions of and entry points to different parts of this PR: 

### S3 mocking
1. This PR creates an [S3 mock](https://github.com/TechAndCheck/zenodotus/blob/206-mock-hypatia-s3/test/aws_s3_mock.rb) class that stands in for `AwsS3Downloader` in tests. 
    1. The mock uses s3 keys to look for matching media filenames. If it finds a matching filename in our `test/mock_media` directory, it returns it. 
3. In `test_helper.rb`, we create a proc that's injected into each individual test file to stub out `AwsS3Downloader` with our `AwsS3Mock`. See [these](https://github.com/TechAndCheck/zenodotus/blob/5b3b5e4e45e5cc9553f5a594ab049e364a1ef972/test/test_helper.rb#L19-L21) lines
4. Within each test (see, e.g., [`InstagramPostTest`](https://github.com/TechAndCheck/zenodotus/blob/206-mock-hypatia-s3/test/models/instagram_post_test.rb)), we create an [s3-mocking block](https://github.com/TechAndCheck/zenodotus/blob/5b3b5e4e45e5cc9553f5a594ab049e364a1ef972/test/models/instagram_post_test.rb#L12-L16) that uses the proc mentioned in (2) to do the actual stubbing. 

### Hypatia mocking
1. This PR creates a [Hypatia mock module](https://github.com/TechAndCheck/zenodotus/blob/206-mock-hypatia-s3/test/hypatia_mock.rb/) stands in for Hypatia in tests. 
    1. The mock creates a hash that matches media post URLs to hypatia responses. When it receives a scrape request for a media post URL, it returns the corresponding hypatia response. 
3. When Minitest imports the mock module into [`test_helper.rb`](https://github.com/TechAndCheck/zenodotus/blob/206-mock-hypatia-s3/test/test_helper.rb), it runs the module's [`construct_typhoeus_hypatia_stub`](construct_typhoeus_hypatia_stub) method, which reroutes Typhoeus requests headed to Hypatia to our mock. 
4. All calls to Hypatia within tests are automatically mocked, so long as we have data for the URL the test tries to scrape from. If a URL isn't caught by the mock, we raise an exception

# Testing instructions
1. `bundle`
2. `rails db:migrate` (pulls in a migration from the base `post-screenshots` branch)
3. `rake test`

# Before merging
1. Update dev README


app/models/sources/facebook_user.rb
```diff
@@ -75,12 +82,11 @@ def self.facebook_user_hash_from_forki_user(forki_user)
         verified:            forki_user["verified"],
         url:                 forki_user["profile_link"],
         profile_image_url:   forki_user["profile_image_url"],
-        profile_image:       File.open(tempfile.path, binmode: true)
+        profile_image:       File.open(profile_image_path, binmode: true)
       }
 
       # It's good practice to unlink the tempfile, even though the next time the GC sees it,
       # it'll be done automatically.
-      tempfile.close!
```
reefdog, line 83: Two quick notes:

1. I have a vague memory of discovering that the path was no longer available after closing the file, even when assigned to a new variable. Which sounds apocryphal so tagging in @cguess, do you recall something similar? If that's the case, we should probably leave this here but I guess make it like `tempfile.close! if tempfile` or something.
2. If we _do_ determine moving the close is safe, we should move the associated comment (L88-89) too.

app/models/sources/instagram_user.rb
```diff
@@ -77,12 +83,9 @@ def self.instagram_user_hash_from_zorki_user(zorki_user)
       profile:             zorki_user["profile"],
       url:                 zorki_user["profile_link"],
       profile_image_url:   zorki_user["profile_image_url"],
-      profile_image:       File.open(tempfile.path, binmode: true)
+      profile_image:       File.open(profile_image_path, binmode: true)
     }
 
-    # It's good practice to unlink the tempfile, even though the next time the GC sees it,
-    # it'll be done automatically.
-    tempfile.close!
```
reefdog, line 85: I'm not a good enough engineer to know if we need to close the file or not, but I do think we should use the same pattern between this file and `facebook_user.rb`.

app/models/sources/youtube_channel.rb
```diff
@@ -59,8 +59,18 @@ def service_id
...
+    if youtube_archiver_channel.has_key?("aws_profile_image_key")
+      profile_image_path = AwsS3Downloader.download_file_in_s3_received_from_hypatia(youtube_archiver_channel["aws_profile_image_key"])
+    else
+      # We create a temp file and write the image data to it, which yea, is dumb,
+      # and there may be a better way to do it, but this works to fix the encoding issues
+      # (basically, when we call `create` later, Rails tries to convert the string into UTF-8
+      # which obviously breaks everything)
+      tempfile = Tempfile.new(binmode: true)
+      tempfile.write(Base64.decode64(youtube_archiver_channel["channel_image_file"]))
+      profile_image_path = tempfile.path
```
reefdog, line 72: See other comments re: not closing this file. Again, I'm ambivalent about whether we do or not (I don't know the consequences), but I do think we should be consistent between the files.

test/controllers/archive_controller_test.rb
```diff
@@ -36,12 +50,16 @@ class ArchiveControllerTest < ActionDispatch::IntegrationTest
...
-    file.close
 
-    scrape = Scrape.create!({ url: "https://www.instagram.com/p/CBcqOkyDDH8/?utm_source=ig_embed", scrape_type: :instagram })
-    callback_response_json = { scrape_id: scrape.id, scrape_result: scrape_result["scrape_result"] }
+    instagram_mocks_file = File.open("test/mock_data/instagram_posts.json")
+    hypatia_mock_response = JSON.parse(instagram_mocks_file.read)["https://www.instagram.com/p/CBcqOkyDDH8/"]
+    instagram_mocks_file.close
+
+    hypatia_mock_response["scrape_result"] = JSON.dump(hypatia_mock_response["scrape_result"])
+    # file.close
```
reefdog, line 59: We can remove this line.

test/controllers/archive_controller_test.rb
```diff
@@ -36,12 +50,16 @@ class ArchiveControllerTest < ActionDispatch::IntegrationTest
     # First we need to construct a callback response. Since we can't really test the callback flow we
     # construct a mock callback. Since doing that by hard coding is actually annoying as hell we first
     # make a forced scrape call to Hypatia, then create it from there
```
reefdog, line 52: Is this comment still an accurate description of what's going on? I can't quite tell based on reading through the code. If so, cool. If not, let's update.

test/hypatia_mock.rb
```diff
@@ -0,0 +1,104 @@
...
+    mock_data_hashes = parse_mock_files(mock_data_filenames, mock_data_directory)
+
+    mock_data_hashes.reduce(:merge) # combine the hashes
+  end
+
+  # Reads the contents of an array of files and returns parsed versions of them
+  #
+  # @!scope module
+  # @params mock_data_filenames: an array of mock data filenames
+  # @params mock_data_directory: the name of the directory mocked data files are stored in
```
reefdog, line 24: I don't think we need colons separating the var names from the description in the @param definition.

Also I hate to point this out but I think it should be `@param` not `@params`? 😬 Looks like that affects most method defs in this file.

test/hypatia_mock.rb
```diff
@@ -0,0 +1,104 @@
...
+  # @return An array of hashes, each containing data parsed from the mock data files
+  def self.parse_mock_files(mock_data_filenames, mock_data_directory)
+    mock_data_filenames.map do |fn|
+      file_path = "#{mock_data_directory}/#{fn}"
```
reefdog, line 28: I'd recommend always using `File.join` to construct the path, to save us from any pain related to accidental opening/trailing slash issues, or weird hierarchy issues, or whatever. (I think these are actually generated from the `Dir` command earlier, rather than human-typed, so way less chance of error, but still would be a good safeguard against future weirdness IMO.)

test/hypatia_mock.rb
```diff
@@ -0,0 +1,104 @@
...
+  # @params mock_data [Hash] A hash mapping URLs to mocked response bodies
+  # @params media_url [String] The media post URL to return a mock Hypatia response for
+  # @return [Typhoeus::Response] A Typhoeus response with a body containing mocked data
+  def self.generate_force_response(mock_data, media_url)
+    mock_body = mock_data[media_url]
+    mock_body = Marshal.load(Marshal.dump(mock_body))  # make a deep copy to avoid mutating the mock_data hash
+
+    # Stringify response body and scrape_result attribute
+    mock_body["scrape_result"] = JSON.dump(mock_body["scrape_result"])
+    mock_body = JSON.dump(mock_body)
```
reefdog, line 49: I think this would double-encode the `["scrape_result"]` content. Just sanity-checking that that's okay!

test/hypatia_mock.rb
```diff
@@ -0,0 +1,104 @@
...
+
+    Typhoeus::Response.new(body: mock_body, code: 200) # Return the appropriate mocked responses
+  end
+
+  # Returns a simple success response, imitating Hypatia's responsess to non-force scrape requests
```
reefdog, line 54: `responsess` → `responses`

test/hypatia_mock.rb
```diff
@@ -0,0 +1,104 @@
...
+
+  # Validates that the input media URL has been mocked and returns a hypatia response with mocked data
+  #
+  # @!scope module
+  # @params request [Typhoeus::Request] An intercepted Typhoeus request to Hypatia
+  # @params mock_data [Hash] A hash mapping URLs to mocked response bodies
+  # @return [Typhoeus::Response] A Typhoeus response with mocked data
+  # def self.load_force_response(request, mock_data)
+  #   generate_mock_response_for_url(mock_data, media_url)
+  # end
```
reefdog, line 72: Do we need to keep this around or can we remove it? (Fine if you have near-term plans for it.)

test/media_sources/youtube_media_source_test.rb
```diff
@@ -2,20 +2,31 @@
...
     end
   end
 
-  def test_initializing_returns_object
+  test "initializing returns an object" do
     assert_not_nil YoutubeMediaSource.new("https://www.youtube.com/watch?v=Df7UtQTFUMQ")
   end
 
-  def test_extracting_creates_youtube_post_object
+  test "extracting create a YoutubePost object" do
```
reefdog, line 29: `create` → `creates`

reefdog (changes requested):
Asa, this is incredible work. Look at this! Look at this!

```
Finished in 35.828105s, 2.9865 runs/s, 8.1221 assertions/s.
107 runs, 291 assertions, 0 failures, 1 errors, 0 skips
Coverage report generated for Minitest to /Users/justin/Projects/duke/zenodotus/coverage. 1099 / 1354 LOC (81.17%) covered.
```

The one error is the callback one you already noted. 👏

I have a bunch of inline comments and suggestions, some of which should be non-controversial (typos and such), some of which might need conversation, and all of which you are welcome to object to entirely.

I do wonder if we should organize all the mock files into a `test/mocks` directory though. I know that would require modifying all the paths, but having the mock files at the root of `test/` feels weird, and if we put them in `test/mocks/` we could also just have `test/mocks/data/` and `test/mocks/media/` too. I leave this up to you though, it's basically just an OCD thing.

👏

oneroyalace:
> I do wonder if we should organize all the mock files into a `test/mocks` directory though. I know that would require modifying all the paths, but having the mock files at the root of `test/` feels weird, and if we put them in `test/mocks/` we could also just have `test/mocks/data/` and `test/mocks/media/` too. I leave this up to you though, it's basically just an OCD thing.

This bothered me a bit, too :). Happy to change it. Luckily, I'll only need to adjust the file paths in `test_helper.rb`. The rest of the tests inherit mock access from that helper file

app/media_sources/facebook_media_source.rb
```diff
@@ -84,7 +85,7 @@ def retrieve_facebook_post
   def retrieve_facebook_post!
     scrape = Scrape.create!({ url: @url, scrape_type: :instagram })
 
-    params = { auth_key: Figaro.env.HYPATIA_AUTH_KEY, url: @url, callback_id: scrape.id, force: true }
+    params = { auth_key: Figaro.env.HYPATIA_AUTH_KEY, url: @url, callback_id: scrape.id, force: "true" }
```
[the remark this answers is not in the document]
oneroyalace: You know, I got confused because I saw that Hypatia checks whether the force parameter is set to `"true"` (string) as opposed to `true` (boolean). But I guess all HTTP query params are stringified, right? I'll change the force param to boolean types across Zenodotus

app/models/sources/facebook_user.rb
oneroyalace: @reefdog, I believe we're safe here because we assign `tempfile.path` to another variable before we call `tempfile.close`. Ruby is a pass-by-value language (at least for immutable data types), so once we've run `variable=tempfile.path`, we don't have to worry about what happens to `tempfile`. 

I'll move the L88-L89 comment as you suggested and keep this thread open in case @cguess has more to say

oneroyalace:
```
[69, 78] in /home/grog/hub/zenodotus/app/models/sources/facebook_user.rb
   69:         # which obviously breaks everything)
   70:         tempfile = Tempfile.new(binmode: true)
   71:         tempfile.write("hey there!")
   72:         profile_image_path = tempfile.path
   73:         byebug
=> 74:         tempfile.close!
   75:         puts "Profile image path: #{profile_image_path}"
   76:         puts "Tempfile.path: #{profile_image_path}"
   77:       end
   78:
(byebug) tempfile.path.object_id
112120
(byebug) profile_image_path.object_id
112140
(byebug) n

[70, 79] in /home/grog/hub/zenodotus/app/models/sources/facebook_user.rb
   70:         tempfile = Tempfile.new(binmode: true)
   71:         tempfile.write("hey there!")
   72:         profile_image_path = tempfile.path
   73:         byebug
   74:         tempfile.close!
=> 75:         puts "Profile image path: #{profile_image_path}"
   76:         puts "Tempfile.path: #{profile_image_path}"
   77:       end
   78:
   79:       hash_to_return = {
(byebug) tempfile.path
nil
(byebug) profile_image_path
"/tmp/20220831-25162-9lcq57"

```
some proofy pudding ^

oneroyalace: Er, @cguess, thoughts on removing the comment altogether?

reefdog: Super, thanks for that. I can't remember _what_ exactly Chris and I were running into then. Carry on!

app/media_sources/facebook_media_source.rb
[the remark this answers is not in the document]
reefdog: Ah, I see. So the existing check was `== "true"`, but that was implicitly relying on HTTP params always being read as strings (which is valid), so you were making it coherent on the param-construction side.

I'm honestly not sure what's best to do here! Obviously we have to keep the check a string check… but should we always compose these values as strings when constructing the query? Taken to its extreme, that would lead us to do things like `query[:quantity] = quantity.to_s` for numbers or something. So I think your last comment is right… as much as possible let's do the "right thing" on each end (true boolean going in, string check coming out).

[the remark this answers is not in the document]
reefdog: BTW this is more a fun bit of 🤔, nothing critical. I fully trust whatever you think is right.

test/controllers/archive_controller_test.rb
oneroyalace: I call upon @cguess. Should I revert my changes to this test (which still fails for me) and skip it or now?  I wanted to find a way for us to remove the `lib/assets/sample_instagram_scrape.json` file, which I _think_ we can now that we have similar files in the `test/mocks/data` directory

test/hypatia_mock.rb
oneroyalace: Eeeeep you're probably right about both of those issues. I'll fix them up

test/hypatia_mock.rb
oneroyalace: Nope! Thanks for the catch

test/hypatia_mock.rb
oneroyalace: Done!

app/models/sources/facebook_user.rb
cguess: Yep, remove it. Closing a file is specifically a Windows thing and necessity anyways (though good practice) since Windows systems have some weird legacy stuff that means two programs can't even *read* a file at the same time. Linux/BSD/POSIX doesn't have this limitation.

cguess (commented):
This is amazing, like awesome, beyond everything. Great work.

Weirdly, I'm getting this error while running tests:
```
/Users/christopher/Repositories/Reporters_Lab/zenodotus/test/controllers/archive_controller_test.rb:7:in `<class:ArchiveControllerTest>': uninitialized constant Minitest::Hooks (NameError)

  include Minitest::Hooks
                  ^^^^^^^
	from /Users/christopher/Repositories/Reporters_Lab/zenodotus/test/controllers/archive_controller_test.rb:5:in `<top (required)>'
	from /Users/christopher/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/bootsnap-1.11.1/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:30:in `require'
	from /Users/christopher/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/bootsnap-1.11.1/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:30:in `require'
```

Other notes to consider:
- Since the `after_all` is the same for every media source, should we consider monkey patching the `ActiveSupport::TestCase` to automatically include that?
- Ditto for `  include Minitest::Hooks`

test/controllers/jobs_tracker_controller_test.rb
```diff
@@ -5,14 +5,14 @@ class JobsTrackerControllerTest < ActionDispatch::IntegrationTest
 
   test "can view jobs" do
     sign_in users(:user1)
-    InstagramMediaSource.extract("https://www.instagram.com/p/CBcqOkyDDH8/?utm_source=ig_embed", false)
+    InstagramMediaSource.extract("https://www.instagram.com/p/CBcqOkyDDH8/", false)
```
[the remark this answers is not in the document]
oneroyalace: They did. Do you think we could use those URLs in Hypatia's `ScraperControllerTest`

oneroyalace:
> * Since the `after_all` is the same for every media source, should we consider monkey patching the `ActiveSupport::TestCase` to automatically include that?
> * Ditto for `  include Minitest::Hooks`
@cguess I thought about that (or creating a new test class our tests could inherit from), but I think we have a few different types of test classes. Some inherit from `ActionDispatch::IntegrationTest`, some inherit from `ActiveSupport::TestCase`, etc

test/controllers/jobs_tracker_controller_test.rb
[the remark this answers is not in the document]
reefdog: I'd advocate for clean URLs everywhere except in tests where you're explicitly testing URL parsing.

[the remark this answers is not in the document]
cguess: Sounds good to me

cguess:
@oneroyalace that's totally fine and makes sense, disregard my idea.

reefdog:
It woooooorks! 🎉

reefdog (approved):
Huzzah!

cguess (approved):
MERGE THIS (great goddamn job)

Issue #270, opened 2022-07-28
Build applicant email confirmation system


reefdog:
_This task is part of #296 and follows the design of #268._

After prospective users have applied for an account but before they are considered a "real" applicant that an admin should evaluate, they must confirm their email address. We'll do that by sending them a confirmation email using our chosen transactional email platform (#281) and either our own email-confirmation library or, if possible, using `Devise:Confirmable`. (The sticky wicket is that these aren't actual "users" yet.)

**Prerequisites:**

- #281
- #303

reefdog:
As a note to self, we considered whether to send mail through SMTP or using Mailgun's API via their [`mailgun-ruby`](https://github.com/mailgun/mailgun-ruby) gem. We decided to use the API for its substantially-better security over SMTP.

The only downside is adding a dependency and slightly increasing the friction of migrating from Mailgun to another platform. but that's not terribly hard:

- Switching from Mailgun SMTP → [Another Platform] SMTP would just mean changing env vars, no code.
- Switching from Mailgun API → [Any SMTP would mean changing a few config settings (remove `action_mailer.mailgun_settings`, add `action_mailer.smtp_settings`) and env vars.
- Switching from Mailgun API → [Another Platform] API is an unknown amount of work and depends on if that other API offers a good Ruby gem with Action Mailer integration.

reefdog:
Delivered by #337 (and #336)! There are some improvements to be made, e.g. improving templates and such, but those can wait for when we're implementing the real design (#267 and beyond).

PR #336, opened 2022-09-07
Stub email confirmation process for applicants

reefdog:
This PR partially implements the email confirmation process for new applicants. The "partial" part is that it does not, in fact, contain _any_ of infrastructure or changes for sending the applicant a confirmation email, but it does:

- Add confirmation-related attributes and methods to the `Applicant` model, using `Devise::Confirmable` as a template (but not actually _using_ that module)
- Add the controller methods and views for guiding the applicant through the process, including actually confirming their application _if_ you know how to magically compose the URL

In the spirit of smaller PRs, I'm submitting this one while filling in the gaps for actually sending the applicant an email.

**Testing the positive path:**
- `rails db:migrate`
- `./bin/dev` (not just `rails s`, because we updated CSS)
- Go to http://localhost:3000/apply/new and submit a new application
- ✅ You should see the ["Next, confirm your email address" screen](https://user-images.githubusercontent.com/4731/188989758-a0add9c3-0358-499f-b606-16a391e8bbfb.png).
- Open the Rails console and generate the confirmation URL for that new applicant `include Rails.application.routes.url_helpers; applicant = Applicant.last; applicant_confirm_url(email: applicant.email, token: applicant.confirmation_token, host: "localhost:3000")`
- Open that URL in a browser
- ✅ You should see the ["Email address confirmed" screen](https://user-images.githubusercontent.com/4731/188990829-549db105-4696-4f72-b050-6e685106a05b.png).

**Testing the error paths:**

There are multiple errors you should attempt to trigger, among them:

- Flip `sent = true` to `sent = false` in [`ApplicantsController#send_confirmation_email`](https://github.com/TechAndCheck/zenodotus/blob/8f72dab31663a4cac383164c51eaf68ee68d779d/app/controllers/applicants_controller.rb#L122) and try submitting a new application. You should see the ["Problem sending confirmation link" message](https://user-images.githubusercontent.com/4731/188992902-449472c0-0a02-468b-99ff-7da36c585c31.png).
- Visit http://localhost:3000/apply/confirm with no parameters (or with junk `email` or `token` params). You should see this we-can-neither-confirm-nor-deny error, ["Email confirmation could not be found"](https://user-images.githubusercontent.com/4731/188992852-29b1439a-5d83-4329-8b8a-310fc11655f4.png).
- Visit the confirmation URL for the applicant you already confirmed. You should see the same "could not be found" message.

Affects #270


app/models/applicant.rb
```diff
@@ -9,10 +9,37 @@ class Applicant < ApplicationRecord
...
+  validates :confirmation_token, uniqueness: true
 
   after_initialize :map_terms_database_values_to_accessor
   before_create :map_terms_accessor_to_database_values
 
+  # Sets the `confirmed_at` attribute to now, which is our proxy for confirmation.
+  # Returns either the boolean result of the `update` or nil if the applicant is already confirmed.
+  sig { returns(T.nilable(T::Boolean)) }
+  def confirm
+    self.update(confirmed_at: Time.now) unless self.confirmed?
```
oneroyalace, line 21: Would it make sense to call `self.update!` and raise an exception if confirmation fails?

oneroyalace (commented):
Code LGTM and your tests pass for me (although I think you meant to write `applicant_confirm_url` instead of `new_applicant_confirm_url` in the pr description)

Few comments:

1. There may be something funny with the CSS on the confirmation page:. The `Return to home button` overlaps with the text above
image
2. For whatever reason, `applicant.confirmed?` didn't update until I restarted the rails console. 

Look forward to peeking at #337 !

reefdog:
> although I think you meant to write `applicant_confirm_url` instead of new_applicant_confirm_url` in the pr description

Oops, good catch! I actually renamed the route after creating the PR and force-pushed the change, but forgot to update the desc. Will fix!

> There may be something funny with the CSS on the confirmation page

Whoops, I forgot that you need to run `./bin/dev` instead of `rails s` to pick up the new CSS. Apologies. Try with `./bin/dev` and see if it looks correct!

> For whatever reason, `applicant.confirmed?` didn't update until I restarted the rails console.

Hrm, I wonder if you need to `applicant.reload!` in between calls?

app/models/applicant.rb
oneroyalace: Hm. Does @joelwluther  need to be alerted when a confirmation fails? If there's no exception, could an application fall through the cracks? I suppose the applicant would still appear on his admin docket.

oneroyalace:
@reefdog you're right. Forgot the `object.reload!` call. That's burned me in the past :(

And yep, perfect! `./bin/dev` works :)

app/models/applicant.rb
reefdog: > Does @joelwluther need to be alerted when a confirmation fails?

Yes! And this is a good point; I intend to notify admins when things fail (e.g., [notifying admins when the notification email doesn't get sent](https://github.com/TechAndCheck/zenodotus/blob/434b68f3c8fe5d585168b6d0c94ace6851595149/app/controllers/applicants_controller.rb#L128), though this isn't implemented yet), but I didn't do that here. I think we should, good catch.

> If there's no exception, could an application fall through the cracks?

This is the point of debate, I think: I can tell the update failed just by looking for a false `update()` response, so is there an advantage to getting an exception instead of just a boolean `false`? Do I get more information to pass along to the admin with the exception? (Again, genuine question, as I don't know!)

> I suppose the applicant would still appear on his admin docket.

The current design model is that admins would be able to see unconfirmed applications, but they'd have to go looking for them; they'd be hidden from initial view. (The idea being, don't fill their inbox with applicants who couldn't be bothered to even confirm their email, or who entered junk addresses, etc.)

oneroyalace (approved):
LGTM

app/controllers/applicants_controller.rb
```diff
@@ -18,21 +18,82 @@ class CreateApplicantParams < T::Struct
...
     TypedParams[CreateApplicantParams].new.extract!(applicant_params)
 
-    @applicant = Applicant.new(applicant_params)
+    # TODO: Throw an error if a user with this email exists, but without confirming existence.
 
-    if @applicant.save
-      redirect_to new_applicant_thanks_path
+    @applicant = Applicant.create(applicant_params.merge(confirmation_token: Devise.friendly_token))
+
+    if @applicant
```
cguess, line 28: Just a style thing, but could we rewrite this without the nested `if else`? Basically move the top level `else` statement up and return early?

app/controllers/applicants_controller.rb
```diff
@@ -18,21 +18,82 @@ class CreateApplicantParams < T::Struct
...
+  # whether because a parameter is missing or we can't match them to an unconfirmed applicant. This
+  # is because we don't want to implicitly reveal the existence or status of any other applicant.
+  sig { void }
+  def confirm
+    # Verify the types of the values provided in the request.
+    # This is wrapped in a begin/rescue because it throws.
+    begin
+      # TODO: Actually use the output of this, if possible.
+      TypedParams[ConfirmApplicantEmailParams].new.extract!(confirm_params)
+    rescue
```
cguess, line 64: Should probably rescue a specific type of error here instead of just eating *everything*

app/controllers/applicants_controller.rb
```diff
@@ -47,4 +108,31 @@ def applicant_params
...
+  def send_confirmation_email
+    # 1. TODO: Actually send email
+    sent = true
+
+    if sent
+      # This shouldn't fail but we don't particularly care if it does.
+      @applicant.update(confirmation_sent_at: Time.now)
+    else
+      notify_of_send_confirmation_error
+      return false
```
cguess, line 129: What if we set "sent" to false here instead and then make line 132 the solo return? Otherwise line 132 is redundant.

app/models/applicant.rb
```diff
@@ -9,10 +9,37 @@ class Applicant < ApplicationRecord
...
+    token: String
+  ).returns(T.nilable(T.self_type)) }
+  def self.find_unconfirmed_by_email_and_token(email:, token:)
+    # Even though tokens are unique (constrained by the database), we want to require the matching
+    # email address, and limit to unconfirmed applicants.
+    self.where({
+      email: email,
+      confirmation_token: token,
+      confirmed_at: nil
+    }).first
```
cguess, line 40: This will raise an exception if nothing is found. Is this what you want?

cguess (approved):
I have a few style questions and one functional one, but other than that consider this approved.

app/controllers/applicants_controller.rb
reefdog: So basically have:

```
unless @applicant
  flash.now…
  render …
end

# and then the send_confirmation_email bit
```

I do get the whole "whittle out error states first, and end with the success state" design overall, but reading the code always requires so many inversions that it just feels less legible to me. I am fine adapting this though, if you have a preference for it! (Truly.)

app/controllers/applicants_controller.rb
reefdog: This whole block gets rewritten in #337, so I'll leave it as-is for this PR, but yeah this was a mix of two design ideas that didn't get resolved, apologies!

app/models/applicant.rb
reefdog: I thought it would as well, but it actually doesn't! `where()` returns an empty array when nothing matches, so `first` is still a valid method.

```
irb(main):001:0> Applicant.where(email: 'asdf')
  Applicant Load (5.0ms)  SELECT "applicants".* FROM "applicants" WHERE "applicants"."email" = $1  [["email", "asdf"]]
=> []                                                                
irb(main):002:0> Applicant.where(email: 'asdf').first
  Applicant Load (0.4ms)  SELECT "applicants".* FROM "applicants" WHERE "applicants"."email" = $1 ORDER BY "applicants"."id" ASC LIMIT $2  [["email", "asdf"], ["LIMIT", 1]]
=> nil                                                                                                   
```

cguess: huh! Thanks!

app/controllers/applicants_controller.rb
reefdog: I took your advice and rewrote this with less nesting and more begging out early. 👍

Issue #273, opened 2022-07-28
Build applicant-management admin system


reefdog:
_This task is part of #296 and follows the design of #268._

Admins need to be able to see applicants on the site, review their applications, and take action on them. The foundation for this was created in #271, but reserved for the console; this task is for designing and building the actual UI for this.

Issue #271, opened 2022-07-28
Build applicant → user pipeline


reefdog:
_This task is part of #296 and follows the design of #268._

Once applicants have confirmed their email address, we need to be able to approve or reject them, along with a reason. This approval/rejection should be communicated back to the applicant. If approved, they should be converted to a proper user, but forced to choose a password upon following the welcome link.

This issue is for that functionality, but the actual admin approval/rejection will be triggered by a Rails console method. (The UI in the site itself will be done in a future task.)

- [x] Approve/reject applications (with reasons)
- [x] Create users from approved applicants
- [x] Allow users to choose password

PR #340, opened 2022-09-09
Add logic to review applicants

reefdog:
This PR adds the ability to review applicants (approve or reject) from the console using the `Applicant#approve` and `Applicant#revoke` methods. The applicant must have already confirmed their email address. Review notes are optional.

⚠️ This is just the first of several steps toward completion of #271, so it doesn't actually create a user yet, nor does it offer an interface in the site for this (coming #273).

**Testing:**
1. 🚨 `bundle install` for a necessary `sorbet-rails` update
2. 🚨 `rails db:migrate`
3. `rails t test/models/applicant_test.rb`
4. Start Rails and Sidekiq, create a new request for access, and confirm its email address
5. `rails c` and find your new applicant (`a = Applicant.last` perhaps)
6. Approve or reject with `a.approve` or `a.reject`
   - Optionally, provide a review note by passing `review_note:` or `review_note_internal:` to those methods
7. ✅ Verify approval/rejection with `a.approved?` and `a.rejected?` or the blanket `a.reviewed?` (true for both), remembering that you may need to `a.reload`

**Testing failure cases:**
- You should **not** be able to review applicants whose emails aren't confirmed (`confirmed_at == nil`).
- You should **not** be able to reject applicants that were already reviewed (`status != nil`).
- You should **not** be able to approve applicants that were already approved (`status == "approved"`).
   - Note the implication of this last one: you **should** be able to approve applicants that were previously rejected. This is intentional.

👋  The easiest way to reset an applicant is just to null out the `status` column manually rather than create a whole new applicant.

Affects #271


app/models/applicant.rb
```diff
@@ -45,6 +50,65 @@ def self.find_unconfirmed_by_email_and_token(email:, token:)
...
+  def reject(
+    review_note: nil,
+    review_note_internal: nil
+  )
+    raise UnconfirmedError unless self.confirmed?
+    raise StatusChangeError if self.reviewed?
+
+    self.update!({
+      status: :rejected,
+      review_note:,
```
oneroyalace, line 96: I believe @cguess prefers explicit key/value syntax for hashes

oneroyalace (approved):
LGTM. Side-note: I can't get emails to send anymore. Do you see any failures on the mailgun side?

app/models/applicant.rb
reefdog: I felt a little bit dirty myself, but I do love these implicit values. @cguess what do you think? (I'm _super_ used to it from JS, where the syntax is even cleaner and doesn't require the trailing colon, but if you want to keep it out of Ruby, I get it.)

reefdog: @cguess [said](https://github.com/TechAndCheck/zenodotus/pull/340#discussion_r968572149):

> I'd prefer to keep it out. I just like explicitness.

Will change!

reefdog:
Oh, for history's sake I should probably note that we determined the sending issue above was just Sidekiq not running locally. (I'd forgotten to note in the PR desc that it needed to be running. It always does, for mail delivery, as we use `deliver_later`.)

Issue #298, opened 2022-08-05
Redesign user settings area


reefdog:
We currently allow users to modify their settings via the Media Vault user area. This should move to the greater FCI area, and get a UX pass.

reefdog:
See the Safari-related UI bugs @cguess noted here: https://github.com/TechAndCheck/zenodotus/pull/343#issuecomment-1245390273

reefdog:
Be sure also to handle #355.

PR #343, opened 2022-09-12
Enable change-of-email confirmation for users

reefdog:
This PR enables email confirmation for users changing their email address. This is a necessary prerequisite for cleanly converting applicants (which require confirmed emails) to users, and is good form besides.

**Testing:**
- Start Rails and Sidekiq
- Log in and go to your user settings screen (http://localhost:3000/account)
- Update your email address to a new one (that you actually have access to)
- ✅ See a flash that tells you to go check your email
- ✅ Verify (in console/database) that your email has not _actually_ changed yet
- Go check your email! Click the confirmation link inside. (Ignore the email design and language. This is Devise boilerplate that we will customize.)
- ✅ Clicking the link should confirm the email change; verify in console/database

Issue #271

### DIFF
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -44,7 +44,7 @@ class AccountsController < ApplicationController
     current_user.email = typed_params.email
     current_user.save
     respond_to do |format|
-      flash.now[:alert] = "Email updated."
+      flash.now[:alert] = "We just sent a confirmation message to the email address you provided. Please check your inbox and follow the confirmation link in the message."
       format.turbo_stream { render turbo_stream: [
         turbo_stream.replace("flash", partial: "layouts/flashes/turbo_flashes", locals: { flash: flash }),
       ] }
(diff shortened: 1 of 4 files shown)


cguess:
The Update Account Details page is failing on Safari. Specifically the text boxes are too small to actually see the text you're typing and the input box "jumps" around after the first input (see screenshots below). Specifically:
- Open page
- Click on "change email" input box
- type something (the hint text will disappear, but no more text will show up)
- click on the form outside the box, the box will jump lower

Screen Shot 2022-09-13 at 9 04 53 AM

I'm going to finish up the rest of the testing on Firefox, but this should be addressed since I'd bet it'll break on iOS and iPadOS as well.

cguess (changes requested):
This all works great except on Safari (the text box is just too small I'd guess). Otherwise it's good to go.

reefdog:
@cguess This PR doesn't change the user settings screens, just the back-end logic upon entering a new email. (Before, the change happened immediately; now, it requires confirmation.)

I'll be redoing the settings screens entirely in #298, can I copy your notes over to that issue and skip them for this PR?

cguess:
Yep go ahead!

PR #344, opened 2022-09-12
Create a user from a given applicant

reefdog:
_This PR ladders off #343._

This PR adds the ability to create a user given an approved applicant. It's step 2 on our way to #271.

⚠️ Note that you won't be able to actually log in as this user just yet (unless you are very good at guessing passwords). The next PR will add the notification email to the user and the ability for them to change their password.

**Testing:**
- Prerequisite: A **new applicant** whose email **has been confirmed,** and which email **doesn't match an existing user**.
- 🚨 `rails db:migrate`
- ✅ `rails t`
- `rails c`
- Find the prerequisite applicant: `a = Applicant.find(id)`
- Create the accompany user: `u = User.create_from_applicant(a)`
- ✅ It should work! (Though see ⚠️ note above.)
- Destroy the user: `u.destroy`
- ✅ The original applicant **should be destroyed** as well. (This is a change from the initial behavior, see conversation below.)

**Test failing cases:**
- Try to create a user from an applicant whose email matches an existing user's: raises `ActiveRecord::RecordInvalid`
- Try to create a user from an applicant that's not confirmed: raises `User::ApplicantNotApprovedError`

Issue #271

### DIFF
--- a/app/models/applicant.rb
+++ b/app/models/applicant.rb
@@ -1,4 +1,7 @@
 class Applicant < ApplicationRecord
+  # This relationship is optional because not all applicants will become users.
+  belongs_to :user, optional: true
+
   # A boolean value for terms acceptance stored in the database isn't good enough for auditing,
   # but a boolean attribute is useful for magic form composition and validation. Thus, we use a
   # non-database-backed attribute until model creation time, when we generate the timestamp and
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -7,6 +7,8 @@ class User < ApplicationRecord
   has_many :image_searches, dependent: :destroy
   has_many :text_searches, dependent: :destroy
 
+  has_one :applicant, dependent: :destroy
+
   # Include default devise modules. Others available are:
   # :timeoutable and :omniauthable
   devise :database_authenticatable, :registerable,
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -19,4 +19,27 @@ class User < ApplicationRecord
   def super_admin?
     self.super_admin
   end
+
+  # Create a User from an approved Applicant.
+  #
+  # BUG: This should be `applicant: Applicant`, but that throws an error.
+  #      We are temporarily using this absurdly permissive sig that works.
+  #      See: https://github.com/TechAndCheck/zenodotus/issues/342
+  sig { params(applicant: Object).returns(User) }
+  def self.create_from_applicant(applicant)
+    raise ApplicantNotApprovedError unless applicant.approved?
+
+    self.create!({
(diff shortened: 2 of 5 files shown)


cguess (approved):
Hrm... why are we not removing the applicant when the user is deleted? Is there a use case where we would use the applicant object when no user is around?

Otherwise, the code and such works great!

reefdog:
We keep the applicant around for auditing history. If you think of the "applicant" instead as an "application" (in the sense of applying for something), I think it makes sense, I just didn't want to use that overloaded word. We should maintain a history of when they applied, what information they provided on their application, who reviewed their application¹, and what reason was given for rejecting or approving them.

That's all information I can see us wanting to reference even after creating a user, and IMO it makes more sense to maintain that info on the applicant object than copy all of it over to the user.

How's that sound?

¹ Technically I'm not yet storing _who_ approved, since approval is happening via the console. As soon as it's happening through the controller, we'll store the associated reviewer's user ID.

reefdog:
> Question: after the user is deleted the applicant is still marked as approved. Is this correct?

Oh, I didn't address this question.

That's a good point, but I think that the answer is yes. This are technically discrete actions: an applicant can be approved without a user being immediately created. Our _process_ will link them, but they are still discrete actions. And so conversely, deleting the associated user shouldn't automatically un-approve the applicant.

More likely, we'll never actually delete users, we'll instead deactivate them. And if we want to wrap that up in a history-rewriting "retroactively reject this applicant" process that deactivates the user and marks the applicant as rejected, we can. But those are both process questions and shouldn't be reflected in a model binding between user deletion and applicant rejection.

cguess:
The only concern is that, for GDPR compliance, if we get asked to delete a user *by* the user, we have to delete *everything* that we have on our service that relates to their identification. So we would deactivate user accounts if we want to kick someone off, but if they request we have to be able to delete everything associated with someone (that's identifiable).

reefdog:
Ooooh, that's a very good point and use case! 🤔

Okay so thinking out loud… the only reason we plan to destroy a user is when the user requests deletion, at which point we should delete their associated applicant record too. _We_ don't intend to ever delete a user, just deactivate, which preserves the applicant record. This argues for leaning on `dependent: :destroy`.

But… it still makes me a little nervous to couple them that way, just because this really isn't a semantic or technical dependency, it's a compliance dependency that I'd _prefer_ to solve by baking applicant-deletion into a special `User#destroy` method, like `User#self_destruct` or something. But then we're basically recreating built-in behavior for the benefit of semantic purity. 🤔

reefdog:
Okay, I'm convinced.

Issue #300, opened 2022-08-05
Add additional MediaVault-specific permissions to users if needed


reefdog:
If there are any specific sub-permissions for Media Vault users, make sure they're settable using the role/permissions system (from #282).

joelwluther:
I think we should have two options here: read-only, for researchers/academics/etc., and then a separate user setting that enables people to submit.

reefdog:
When you say submit, does that just mean archiving a new piece of media, or submitting a Media Review too? And would this archived media be part of the public corpus or just a private "my own archive" area?

Chris and I have discussed the possible malicious consequences of letting people add to the public corpus, and our liability in all cases of letting users have us scrape things. We'll want to make sure we have really good moderation capabilities.

(It's good to know about these permissions in advance though, thanks! I'll be sure we're robust enough to support this.)

joelwluther:
Mmm I meant specifically archiving media -- we're supporting our three FactStream partners in creating MediaReview; I don't think we want to expand that. Others can use the FCMT or json-ld embeds, as they've done with ClaimReview.

Actually, thinking through it, if the archiving happens at the point of creating the MediaReview (and then us ingesting the data and scraping), we don't really need the separate permissions. Though we do want to be clear that Vault is opt-in, Insights is opt-out.

reefdog:
> if the archiving happens at the point of creating the MediaReview (and then us ingesting the data and scraping), we don't really need the separate permissions

Yup, @cguess can verify, but the architecture we're building toward is automatically scraping new MediaReviews to the Vault at time of entry. May even already be happening.

> Though we do want to be clear that Vault is opt-in, Insights is opt-out.

Correct! In fact, right now the idea is that users don't even opt in to Vault; they're just quietly given permission by an admin if the admin recognizes that a given applicant or user should have it. This is based on some conversation on our last group meeting about keeping Vault's visibility low.

There is a middle ground: don't pitch Vault on the public FCI site, but perhaps let all confirmed / logged-in FCI users see a Vault pitch page that gives them a vague description and lets them request access. It just depends on how much visibility we want to provide.

reefdog:
Although also, we don't actually have a concept of "opting out" of Insights currently. If you have an account, you have access to the data. The architecture currently has no path for being an authenticated user who has opted out of Insights access.

joelwluther:
[...]

Issue #282, opened 2022-07-29
Evaluate and choose role/permission gem(s)


reefdog:
We currently plan to have **two roles** (`Admin` and `User`) and only **one permission** (`hasMediaVaultAccess`). However, our solution should assume neither of those are constant.

- ✅ **Authentication**: We're using [Devise](https://github.com/plataformatec/devise).
- ⏳ **Authorization** candidates:
   - [CanCanCan](https://github.com/CanCanCommunity/cancancan)
   - [Pundit](https://github.com/varvet/pundit)
- ⏳ **Role definition/management** candidates:
   - [Rolify](https://github.com/RolifyCommunity/rolify)

reefdog:
I've dived through the documentation for the suggestions above, and done a little more surveying of options. There aren't a lot! I've also thought more about this and our **actual current needs:**

- All our users have access to the FCI data
- Some of those users may have access to Media Vault
- All of our admins can do everything

Is there anything else even on the horizon? Perhaps… being able to upload your own things to Media Vault might be something we restrict to certain users? (I don't think we ever intend for people to be able to upload to the shared corpus, so this would just be the ability to have your own private archive.)

Given the simplicity of our current set of permissions, and that they are all _application_ permissions (not _resource_ permissions) I think we should strongly consider _not_ adding the complexity of a gem. This would be much easier to add just as attributes on the `User` model.

**Big caveat to all of this:** We're not going to be using the "organization" concept that's hiding in the current codebase. If we were, we would _definitely_ be considering a gem, since it would handle precisely this sort of scoped/relative authorizations.

@cguess @oneroyalace Would love have chat for a few minutes with y'all about this before moving forward!

reefdog:
Update: Okay, I actually dived a little _more_ into the documentation for Rolify (specifically) and thought about it, and… I think Rolify on its own is a sensible middle ground that would minimize future disruption if we wanted to get more complex with our permission structure.

So I'd like to choose `Rolify` on its own without any authorization gem; in the future we can insert one as needed (I think).

reefdog:
Got sign-off from @cguess to start with Rolify sans CanCanCan.

Issue #318, opened 2022-08-25
Remove Twitter in Lieu of Hypatia


cguess:
We've moved Twitter to Hypatia, remove it from the code base.

PR #349, opened 2022-09-20
Add user roles

### LINKED ISSUE #299: Add Rolify and roles to user model
With #271 in existence, and based on the decisions in #282, we should add [Rolify](https://github.com/RolifyCommunity/rolify) to the User model, and add roles such as:

- New user that hasn't completed setup (default)
- Insights user
- MediaVault user
- Admin user

Users should be able to have multiple roles simultaneously, although the only two roles we currently plan to be active simultaneously is "Insights + MediaVault".

reefdog:
This PR adds the concept of user roles to the site, and converts our determination of "what is an admin?" to use it.

It adds the Rolify gem (as determined in #282), which lets us add roles ad-hoc both "globally" (applies to the user everywhere) and per-resource (e.g. "user can moderate this particular archive item"), although we're only using the global roles. Roles are non-exclusive; a user can have multiple roles at once.

We currently have four roles:

- `new_user` is assigned during `User` creation, and it semantically means "this user has not yet completed their setup process". This role is removed when the user follows the link in their email and chooses their password.
- `insights_user` is also assigned during `User` creation, and it semantically means "this user is approved to use the Insights site". This is a little bit unnecessary since all users can use the site, but I think it's good to be explicit and look to a future where this might not be the case (e.g., Vault-only users).
- `media_vault_user` are users who have the right to use Media Vault
- `admin` are users who are administrators of the site

The only role that is currently used for any gatekeeping is `admin`, which replaces the `super_admin` attribute on the `User` model. Rolify provides dynamic methods (defined during app boot and during new role creation, not through "method missing") for each role, e.g. `is_admin?` or `is_new_user?`, and we make use of those.

⚠️ Note that this PR adds no _new_ functionality; it is just a clean swap of the old system with roles. Future changes will actually use the roles for more gatekeeping (e.g., #300).

**Testing:**
- 🚨 `bundle install`
- 🚨 `rails db:reset` ← you really should do a full reset if possible, to get all the new seeds
- ✅ `rails t` of course
- Run the app and login with the standard `admin@example.com` and `user@example.com`
- ✅ Nothing should be different from before! Login should work, only admins should get the Jobs page in their user menu, etc.

Closes #299

### DIFF
--- a/config/initializers/rolify.rb
+++ b/config/initializers/rolify.rb
@@ -1,0 +1,10 @@
+Rolify.configure do |config|
+  # By default ORM adapter is ActiveRecord. uncomment to use mongoid
+  # config.use_mongoid
+
+  # Dynamic shortcuts for User class (user.is_admin? like methods). Default is: false
+  config.use_dynamic_shortcuts
+
+  # Configuration to remove roles from database once the last resource is removed. Default is: true
+  # config.remove_role_if_empty = false
+end
--- a/config/initializers/rolify.rb
+++ b/config/initializers/rolify.rb
@@ -3,8 +3,14 @@ Rolify.configure do |config|
   # config.use_mongoid
 
   # Dynamic shortcuts for User class (user.is_admin? like methods). Default is: false
+  #
+  # Enabled because these are convenient methods, and according to the Rolify documentation they
+  # are generated at boot time (and when `add_role` is run), so shouldn't hurt performance.
   config.use_dynamic_shortcuts
 
   # Configuration to remove roles from database once the last resource is removed. Default is: true
-  # config.remove_role_if_empty = false
+  #
+  # Toggled to false because we have well-defined user roles that we don't want removed, even if
+  # the last user using them is deleted.
+  config.remove_role_if_empty = false
 end
--- a/db/migrate/20220919214503_convert_admins_to_rolify.rb
+++ b/db/migrate/20220919214503_convert_admins_to_rolify.rb
@@ -1,0 +1,15 @@
+class ConvertAdminsToRolify < ActiveRecord::Migration[7.0]
+  def up
+    User.where(super_admin: true).each { |admin| admin.add_role :admin }
+
+    remove_column :users, :super_admin
+  end
+  def down
+    add_column :users, :super_admin, :boolean, default: false
+
+    User.with_role(:admin).each do |admin|
+      admin.update super_admin: true
+      admin.remove_role :admin
+    end
+  end
+end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -19,7 +19,7 @@ AllCops:
 # Join tables don't really need timestamps
 Rails/CreateTableWithTimestamps:
   Exclude:
-    # - 'db/migrate/20180128231930_create_organizations_and_events.rb'
+    - db/migrate/20220919194501_rolify_create_roles.rb
 
 # Rails generates this file
 Style/BlockComments:
@@ -64,6 +64,10 @@ Rails/HasManyOrHasOneDependent:
 
 Rails/HasAndBelongsToMany:
   Enabled: true
+  Exclude:
+    # Rolify uses HABTM. Despite a decade of the community attempting to implement
+    # `has_many: through`, it still struggles mightily with it. Let's make an exception.
+    - app/models/role.rb
 
 Style/NumericPredicate:
   Enabled: true
--- a/Gemfile
+++ b/Gemfile
@@ -178,3 +178,6 @@ gem "country_select", "~> 8.0"
 
 # Used for sending email through Mailgun
 gem "mailgun-ruby", "~> 1.2"
+
+# Rolify is used for user roles
+gem "rolify", "~> 6.0"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -326,6 +326,7 @@ GEM
       mime-types (>= 1.16, < 4.0)
       netrc (~> 0.8)
     rexml (3.2.5)
+    rolify (6.0.0)
     rubocop (1.28.1)
       parallel (~> 1.10)
       parser (>= 3.1.0.0)
@@ -506,6 +507,7 @@ DEPENDENCIES
   rails (~> 7.0.2.3)
   rake
   redis (~> 4.0)
+  rolify (~> 6.0)
   rubocop
   rubocop-minitest
   rubocop-performance
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -87,6 +87,8 @@ class AccountsController < ApplicationController
     raise InvalidTokenError if @user.new_record?
     raise InvalidUpdatePasswordError if typed_params.password.blank? || @user.invalid?
 
+    @user.remove_role :new_user
+
     sign_in @user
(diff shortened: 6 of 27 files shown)


db/migrate/20220919214503_convert_admins_to_rolify.rb
```diff
@@ -0,0 +1,15 @@
...
+  end
+  def down
+    add_column :users, :super_admin, :boolean, default: false
+
+    User.with_role(:admin).each do |admin|
+      admin.update super_admin: true
+      admin.remove_role :admin
+    end
+  end
+end
```
reefdog, line 15: I wanted to draw attention to this destructive migration, which seeks to migrate the old super-admin data to the new.

reefdog:
This PR is tripping up Rubocop with two offenses:

```
Offenses:

app/models/role.rb:2:3: C: Rails/HasAndBelongsToMany: Prefer has_many :through to has_and_belongs_to_many.
  has_and_belongs_to_many :users, join_table: :users_roles
  ^^^^^^^^^^^^^^^^^^^^^^^
db/migrate/20220919194501_rolify_create_roles.rb:10:5: C: Rails/CreateTableWithTimestamps: Add timestamps when creating a new table.
    create_table "users_roles", id: false do |t| ...
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

I'd like to create the following two exceptions:

```yml
Rails/HasAndBelongsToMany:
  Enabled: true
  Exclude:
    # Rolify uses HABTM. Despite a decade of the community attempting to implement
    # `has_many: through`, it still struggles mightily with it. Let's make an exception.
    - app/models/role.rb
```

Rolify uses HABTM and, judging by the literal decade of issues and PRs attempting to implement `has_many :through`, the community hasn't been able to make it work correctly. (See: [PR#181](https://github.com/RolifyCommunity/rolify/pull/181), [PR#571](https://github.com/RolifyCommunity/rolify/pull/571), [PR#583](583), [#318](https://github.com/RolifyCommunity/rolify/issues/318), [#411](https://github.com/RolifyCommunity/rolify/issues/411)…) I think we're going to really twist ourselves in knots trying to get around it.

We should either ditch Rolify over this, or make an exception for this case.

```yml
# Join tables don't really need timestamps
Rails/CreateTableWithTimestamps:
  Exclude:
    - db/migrate/20220919194501_rolify_create_roles.rb
```

This one's pretty straightforward. We _could_ create timestamps, but as the comment says — which I didn't add, it was already in our Rubocop file — a true dumb join table doesn't really need them.

cguess:
Running tests gives me two errors (below):
```

Error:
ApiKeyTest#test_deleting_a_user_deletes_all_of_its_api_keys:
StandardError: No fixture named 'user3' found for fixture set 'users'
    test/models/api_key_test.rb:18:in `block in <class:ApiKeyTest>'


rails test test/models/api_key_test.rb:17

..........E

Error:
ArchiveItemTest#test_destroying_a_user_resets_the_submitter_id_of_ArchiveItems_it_created:
StandardError: No fixture named 'user3' found for fixture set 'users'
    test/models/archive_item_test.rb:16:in `block in <class:ArchiveItemTest>'
    test/models/archive_item_test.rb:11:in `block in around'
    test/models/archive_item_test.rb:10:in `around'


rails test test/models/archive_item_test.rb:15
```

cguess:
> This PR is tripping up Rubocop with two offenses:
> 
> ```
> Offenses:
> 
> app/models/role.rb:2:3: C: Rails/HasAndBelongsToMany: Prefer has_many :through to has_and_belongs_to_many.
>   has_and_belongs_to_many :users, join_table: :users_roles
>   ^^^^^^^^^^^^^^^^^^^^^^^
> db/migrate/20220919194501_rolify_create_roles.rb:10:5: C: Rails/CreateTableWithTimestamps: Add timestamps when creating a new table.
>     create_table "users_roles", id: false do |t| ...
>     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> ```
> 
> I'd like to create the following two exceptions:
> 
> ```yaml
> Rails/HasAndBelongsToMany:
>   Enabled: true
>   Exclude:
>     # Rolify uses HABTM. Despite a decade of the community attempting to implement
>     # `has_many: through`, it still struggles mightily with it. Let's make an exception.
>     - app/models/role.rb
> ```
> 
> Rolify uses HABTM and, judging by the literal decade of issues and PRs attempting to implement `has_many :through`, the community hasn't been able to make it work correctly. (See: [PR#181](https://github.com/RolifyCommunity/rolify/pull/181), [PR#571](https://github.com/RolifyCommunity/rolify/pull/571), [PR#583](583), [#318](https://github.com/RolifyCommunity/rolify/issues/318), [#411](https://github.com/RolifyCommunity/rolify/issues/411)…) I think we're going to really twist ourselves in knots trying to get around it.
> 
> We should either ditch Rolify over this, or make an exception for this case.
> 
> ```yaml
> # Join tables don't really need timestamps
> Rails/CreateTableWithTimestamps:
>   Exclude:
>     - db/migrate/20220919194501_rolify_create_roles.rb
> ```
> 
> This one's pretty straightforward. We _could_ create timestamps, but as [that preexisting comment says](https://github.com/TechAndCheck/zenodotus/blob/78a4e9857811525ac0c10a352ff3f9c06b2e3932/.rubocop.yml#L19-L22), a true dumb join table doesn't really need them.

Do it, sounds good to me!

reefdog:
> Running tests gives me two errors (below):

😱 You're right! I'm not sure how I missed this, apologies. Didn't fully complete the references to the new users fixtures. Fixed and force-pushed!

> > [Suggested Rubocop changes]
>
> Do it, sounds good to me!

Done! This checks clean now.

Could I get a re-review @cguess?

config/initializers/rolify.rb
```diff
@@ -0,0 +1,10 @@
+Rolify.configure do |config|
+  # By default ORM adapter is ActiveRecord. uncomment to use mongoid
+  # config.use_mongoid
+
+  # Dynamic shortcuts for User class (user.is_admin? like methods). Default is: false
+  config.use_dynamic_shortcuts
+
+  # Configuration to remove roles from database once the last resource is removed. Default is: true
+  # config.remove_role_if_empty = false
```
cguess, line 9: Should we turn this off? Obviously we don't think we'll ever have no users, but seems like we wouldn't want the roles removed if we do accidentally drop 'em all.

cguess (approved):
One little question, but otherwise you're good to go.

config/initializers/rolify.rb
reefdog: This is a good question. By design, Rolify's role-creation is dynamic. We don't actually seed the database in advance, or even define valid roles anywhere. We just say `user.add_role :foo` and the `:foo` role gets created if needed (and some helper methods like `is_foo?` automatically defined).

I think you're on to something. Perhaps we should (1) go ahead and add the known roles to the seeds file, and (2) toggle that config setting.

Thanks for the catch!

reefdog: You know… this _might_ only apply when roles have been scoped to a specific resource. In which case we probably _would_ want them deleted. Still, we aren't using using that, so let's enable it and deal with it if/when it becomes a problem.

cguess (approved):

PR #350, opened 2022-09-20
Improve the must_be_logged_out redirection

reefdog:
I wasn’t aware of `redirect_back_or_to` when I wrote my convoluted `ApplicationController#must_be_logged_out` method, despite it being used literally in that same file. Fun!

This quick fix makes use of it. It is caught by a test (`rails t test/controllers/accounts_controller_test`), so run that and make sure it's clean.

There's no accompanying issue, it's just a quick thing I noticed.


cguess (approved):
Nice catch.

PR #357, opened 2022-09-20
Update Rubocop to Ruby 3.1.2

cguess:
This updates Rubocopp so it's ready for the newest version of Ruby. In the process of that it also bumps Ruby from 3.1.1 to 3.1.2.

Here we also set the new JS-style hash syntax to cause an error in Rubocop. Personally I'm just not a fan of it, it makes refactoring more difficult and you have to track down variables left and right when debugging.

After all this we run `rubocop -A` which fixes a number of auto-correctable style issues.

To test:
- Make sure you're running Ruby 3.1.2 `ruby -v`
- `bundle install`
- `rubocop` to check everything
- `rails t` to test

Thanks to @reefdog for bringing this to my attention.


app/models/applicant.rb
```diff
@@ -96,8 +96,8 @@ def reject(
 
     self.update!({
       status: :rejected,
-      review_note:,
-      review_note_internal:,
+      review_note: review_note,
+      review_note_internal: review_note_internal,
```
reefdog, line 100: Darn, you found it.

reefdog (approved):
Looks good! FWIW remember that I had to run `bundle exec rubocop` earlier or else I was getting different results, so in future the instructions might want to mention that. (The results are the same now, but may not as system and repo diverge again.)

🚀

PR #359, opened 2022-09-21
Fix authentication on all controller methods

### LINKED ISSUE #166: Make sure routes/actions are locked down correctly
I noticed some funny behavior when clicking around the site while logged in and not logged in:

**While logged in:**
- Everything seems to work fine, except the "All Collected Media" [placeholder](https://github.com/TechAndCheck/zenodotus/blob/3e0263678a4fd48cd065629d55ce8b5a33ac47e6/app/views/layouts/_header.html.erb#L6) (which goes for everyone of course).

**While not logged in:**
- Clicking "Zenodotus" (root) spawns the login flash, which is a little weird.
- "Text Search" is visible, but clicking it spawns the login flash. (It should just be hidden, right?)
- 🚨 The following pages don't appear on the navbar, but their pages are accessible and usable if you visit them manually:
   - `/image_search`
   - `/twitter_users/*`

So I think the things we need to do:
- [x] Hide Text Search (and maybe All Collected Media?) from unauthenticated users
- [x] Make sure the `image_search` and `twitter_users` routes require authentication
- [x] 🤔 about whether we have a fundamental security model problem if we missed those (should our route authentication be opt-out instead of opt-in?)

reefdog:
This PR locks down all routes that need locking down, ensures the ones that need to be open are open, and adds tests to catch as much of this as possible.

I opened #166 after noticing that I could hit pages I wasn't supposed to be able to, such as a Twitter user page while logged out, or a Jobs status page while logged in as a user (not admin). Now, all pages should be locked down correctly.

It looks like a huge PR, but a ton of it are just new tests. That said, a lot of controllers were touched (and unused ones removed), so feel free to comb through.

⚠️ Note that the commits are organized sensibly, so it might be preferable to go through each commit individually since the description may provide relevant change context.

**Testing:**
- `rails t` obviously
- I highly recommend doing a full login, poke around, archive URL, etc. as multiple types of users. Notify me of anything newly weird.

Closes #166

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -83,7 +83,7 @@ protected
     authenticate_user!
 
     unless current_user.is_admin?
-      redirect_back_or_to "/", allow_other_host: false, alert: "You must be a super user/admin to access this page."
+      redirect_back_or_to "/", allow_other_host: false, alert: "You don’t have permission to access that page."
     end
   end
 end
--- a/app/controllers/archive_controller.rb
+++ b/app/controllers/archive_controller.rb
@@ -1,10 +1,9 @@
 # typed: strict
 
 class ArchiveController < ApplicationController
-  # It's the index, list all the archived items
-
   before_action :authenticate_user!, except: :scrape_result_callback
 
+  # It's the index, list all the archived items
   sig { void }
   def index
     respond_to do | format |
--- a/app/controllers/facebook_users_controller.rb
+++ b/app/controllers/facebook_users_controller.rb
@@ -1,4 +1,8 @@
+# typed: strict
+
 class FacebookUsersController < ApplicationController
+  before_action :authenticate_user!
+
   sig { void }
   def show
     @facebook_user = Sources::FacebookUser.find(params[:id])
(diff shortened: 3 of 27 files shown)


cguess:
First thing I noticed is this Flash message shows up as soon as I log in, even on a fresh login.

Screen Shot 2022-09-22 at 10 48 26 AM

cguess (approved):
This looks great. There's one bug when logging in mentioned above. but once that's fixed consider approved.

cguess:
the flash message is indeed a regression, this is what `master` looks like
Screen Shot 2022-09-22 at 11 09 00 AM

reefdog:
@cguess Confirmed on my end! I've traced it back to [here](https://github.com/TechAndCheck/zenodotus/pull/359/commits/73f301a5e7672118c906c1f9d0fab34ba6071df3#diff-513a92c45a3a5c14210895bd4082fb0d9b6a33dea5209ff25362f03d2a1f300bR6), which was an attempt to prevent logging in while already logged in. Unsure what race condition or whatever is triggering this, but I'm going to reduce this down to just `:new`. It seems fine to stop users from accessing the login screen while logged in, and if someone wants to force a call to the login handler… fine?

Thanks for finding this! I'm going to patch and push.

Issue #360, opened 2022-09-21
Manual URL archiving no longer works


reefdog:
Ever since we added automatic importing of FactStream media from MediaReviews, manual archiving fails because there's no MR for it to be associated with:

```
Completed 500 Internal Server Error in 1930ms (ActiveRecord: 78.3ms | Allocations: 141645)

Scrape::MediaReviewItemNotFoundError (Media review with url: https://www.youtube.com/watch?v=Df7UtQTFUMQ for scrape fulfillment not found.):

app/models/scrape.rb:35:in `fulfill'
app/controllers/archive_controller.rb:98:in `scrape_result_callback'
```

PR #361, opened 2022-09-21
Use the non-`ActiveJob` flow when the "add to archive" button is clicked

oneroyalace:
Our manual archive button currently executes the `ActiveJob` flow, which expects that:
1. An orphaned `MediaReview` item will be created
2. We'll scrape Hypatia and create an `ArchiveItem`
3. We'll attach the orphaned `MediaReview` to the newly created `ArchiveItem`

But step 1 only happens if we either manually create a `MediaReview` item or receive one from FactStream (which happens in production).

Since no one but the dev team uses the "add to archive button", we can change it to skip the `ActiveJob` MediaReview fulfillment flow and just create an `ArchiveItem`.

Closes https://github.com/TechAndCheck/zenodotus/issues/360

### DIFF
--- a/app/controllers/archive_controller.rb
+++ b/app/controllers/archive_controller.rb
@@ -36,7 +36,7 @@ class ArchiveController < ApplicationController
     url = typed_params.url_to_archive
     object_model = ArchiveItem.model_for_url(url)
     begin
-      object_model.create_from_url(url, current_user)
+      object_model.create_from_url!(url, current_user)
     rescue StandardError => e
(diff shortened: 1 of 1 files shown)


reefdog (approved):
🚀

reefdog:
One consequence of this I did notice but failed to mention is that the Archive URL modal now stays open until either manually dismissed or the scrape completes. Since this is just a dev feature that's fine, but I wanted to note it in case there was a quick fix for that, too.

PR #362, opened 2022-09-21
Lowercase applicant emails for better matching against users

### LINKED ISSUE #356: Applicants can be created with same email as users
We intended to prevent applicants from being created with the same email address as an existing user:

https://github.com/TechAndCheck/zenodotus/blob/797c4ea1f82825a3dccd90187c764eca549c55c9/app/controllers/applicants_controller.rb#L34-L36

However, because this is an exact check when it should be a case-insensitive check, it can be gamed simply by entering an email address with different casing from the user. 😬 

Users created through Rails always have lowercase email addresses. We should lowercase `Applicant` email before saving, and also add integration tests that catch this.

reefdog:
_This PR ladders off #359._

This PR closes a loophole that would let people apply using an existing users' email address by changing the casing. (Of course, they'd not have been able to close the loop on becoming a user since they'd hit our uniqueness restrictions during the user-creation process, but this stops them a lot earlier.)

Note that it does the downcasing in two places:

1. As part of a `before_save` action on the model itself; this keeps it out of the database entirely. I could have done this via `after_initialize` so that `Applicant.new` would have it, but I wanted that to represent the user's input until the actual save point.
2. As part of the check for existing users in the controller action. Here I could have changed the casing while passing the params to `Applicant.new`, but that violated the controller-model responsibility split IMO (and would have meant this was the _only_ place the email was downcased).

**Testing:**
- On `master`, apply for an account using `ADMIN@example.com` (or some other cased-differently email currently in your **Users** table)
- ❌ The application should be allowed through. Bad!
- Switch to this branch and repeat the exercise.
- ✅ You should get booted back to the application page with a vague error about being unable to complete. (The language here predates this PR and is intentionally vague so as not to confirm the existence of an account.)

Closes #356

### DIFF
--- a/app/controllers/applicants_controller.rb
+++ b/app/controllers/applicants_controller.rb
@@ -28,8 +28,11 @@ class ApplicantsController < ApplicationController
       generic_create_error && return
     end
 
-    params_with_token = applicant_params.merge(confirmation_token: Devise.friendly_token)
-    @applicant = Applicant.new(params_with_token)
+    decorated_params = applicant_params.merge({
+      # Add the confirmation token the applicant uses to confirm their email address
+      confirmation_token: Devise.friendly_token,
+    })
+    @applicant = Applicant.new(decorated_params)
 
     existing_user = User.readonly.find_by(email: @applicant[:email].downcase)
     # We intentionally return a generic error to avoid leaking the existence of the user.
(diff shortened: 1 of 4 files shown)


oneroyalace (approved):
LGTM

PR #368, opened 2022-09-26
Simplify author JSON export

### LINKED ISSUE #235: Downloading user data shouldn't change URL
**Steps to reproduce:**
1. Go to a user page (e.g., `http://localhost:3000/twitter_users/{id}`)
2. Click the download arrow next to "Archived Media Items"
3. Observe that your file downloads, but also…
4. ❌ That the URL now has `/download` appended to it.

**What should happen:** Clicking the download link should generate and download the file, but without changing the URL.

reefdog:
This PR cleans up how we're generating the JSON export file for our authors. Previously, it was a separate route from the `show` action with its own nearly-identical-but-not-quite query.

Now, it's all within `show` and performs the same query, but customizes the generation based on format requested (HTML v. JSON). It also removes the now-unused routes and adds tests of the JSON generation.

This came out of looking at #235, which was already fixed, but this makes it even more fixed. Fixeder.

**Testing:**
- You may have to `bundle install` depending on previous setup
- `rails t` of course
- Run the app and go to the author pages for each type (Twitter, Instagram, Facebook, and YouTube)
- ✅ Make sure the JSON download button… downloads JSON!

Closes #235

### DIFF
--- a/app/controllers/facebook_users_controller.rb
+++ b/app/controllers/facebook_users_controller.rb
@@ -9,19 +9,14 @@ class FacebookUsersController < ApplicationController
     facebook_post_archive_items = ArchiveItem.includes(archivable_item: [:author, :images, :videos])
                                              .where(archivable_item_type: "Sources::FacebookPost")
     facebook_posts_by_user = facebook_post_archive_items.select { |t| t.facebook_post.author == @facebook_user }
-    @archive_items = facebook_posts_by_user
-  end
 
-  # Exports all media items created by the currently viewed Facebook user to a JSON file
-  sig { void }
-  def export_facebook_user_data
-    facebook_post_archive_items = ArchiveItem.includes(archivable_item: [:author])
-                                             .where(archivable_item_type: "Sources::FacebookPost")
-    facebook_user = Sources::FacebookUser.find(params[:id])
-    facebook_posts_by_user = facebook_post_archive_items.select { |t| t.facebook_post.author == facebook_user }
-    facebook_post_archive_json = ArchiveItem.prune_archive_items(facebook_posts_by_user)
-    send_data facebook_post_archive_json,
-              type: "application/json; header=present",
-              disposition: "attachment; filename=#{facebook_user.name.parameterize(separator: '_')}_facebook_archive.json"
+    respond_to do |format|
+      format.html { @archive_items = facebook_posts_by_user }
+      format.json do
+        send_data ArchiveItem.generate_pruned_json(facebook_posts_by_user),
+          type: "application/json",
+          filename: "#{@facebook_user.name.parameterize(separator: '_')}_facebook_archive.json"
+      end
+    end
   end
 end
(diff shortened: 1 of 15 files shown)


app/controllers/facebook_users_controller.rb
```diff
@@ -9,19 +9,21 @@ def show
...
-    facebook_user = Sources::FacebookUser.find(params[:id])
-    facebook_posts_by_user = facebook_post_archive_items.select { |t| t.facebook_post.author == facebook_user }
-    facebook_post_archive_json = ArchiveItem.prune_archive_items(facebook_posts_by_user)
-    send_data facebook_post_archive_json,
-              type: "application/json; header=present",
-              disposition: "attachment; filename=#{facebook_user.name.parameterize(separator: '_')}_facebook_archive.json"
+private
+
+  sig { params(archive: T.nilable(T::Array[ArchiveItem])).returns(String) }
+  def generate_export_json(archive)
```
cguess, line 26: Question here: Since this function, which is added to all the models, only is a single line, is there a reason `ArchiveItem.prune_archive_items(archive)` wasn't just used instead of an entire function?

app/controllers/facebook_users_controller.rb
```diff
@@ -9,19 +9,21 @@ def show
...
+    ArchiveItem.prune_archive_items(archive)
```
cguess, line 27: I know this was implemented before this PR, probably by me, but right now we're outputting the JSON file as a single line. Question for the room: should we instead pretty print it so that the output is readable by a human? It would just be a single line change to use this function instead: https://ruby-doc.org/stdlib-2.6.3/libdoc/json/rdoc/JSON.html#method-i-pretty_generate

cguess (approved):
I too wasn't able to check every function type, but I think it's good enough to test one since the implementation is the same across the board.

Consider this approved on answering of the two questions I put in.

app/controllers/facebook_users_controller.rb
reefdog: I think pretty-printing is a great idea.

app/controllers/facebook_users_controller.rb
reefdog: That's a good point. My reason for doing it here was to make what was happening more explicit without being so high-touch as to rename `ArchiveItem.prune_archive_items`. But maybe that would have been the better suggestion for me to have made: rename `ArchiveItem.prune_archive_items` to `ArchiveItem.generate_pruned_json` or whatever, and then eliminate the gofer method.

Want me to implement the change? Happy to.

reefdog: BTW, I also considered (and still am) whether, since these are nearly identical, if would make sense to do some meta-programming here with a shared base controller and each author type controller just configures it (with the target model names, attribute chains, etc.). But that felt like maybe a bit too much abstracting until we have even more redundant code?

app/controllers/facebook_users_controller.rb
reefdog: Wrote it up as an issue: https://github.com/TechAndCheck/zenodotus/issues/370

app/controllers/facebook_users_controller.rb
reefdog: @cguess Any objections to me renaming the class method?

cguess: no objections, go ahead

PR #373, opened 2022-09-27
Update `MediaReview` model

### LINKED ISSUE #369: Update `MediaReview` model with Schema.org attributes

oneroyalace:
This PR adds several attributes from https://schema.org/MediaReview to the `MediaReview` model. Most significantly, it introduces the idea of an `associatedClaimReview` to Zenodotus. To represent that attributes, it creates a new `ClaimReview` model, which `belongs_to MediaReview`. 

This PR also updates our `claim-review-schema.json` JSON schema file (renamed to `media-review-schema.json`) to be in accord with the changes mentioned above

# Before merging

- [x] ~@cguess I'll ping you in our devshop channel to ask about how to grab those `MediaReview` JSONs.~ Done!
- [x] ~Depending on what I see in (1), I may remove [these](https://github.com/TechAndCheck/zenodotus/blob/58294f22ca64a6c33093e049025a21a7424b58e9/app/models/archive_item.rb#L24-L28) lines from `archive_item.rb`. The schema.org schema suggests we don't need them and can access a top-level `originalMediaUrl` attribute.~ nvm.

# Testing instructions
1. `rails db:migrate`
1. `rails db:reset` (refresh seeds)
3.  `rake`

Closes #369


reefdog:
Initial code-scan looks good. @oneroyalace could you rebase this from `master` to re-apply your changes to the ingest controller (and the archive/ingest controller tests), and then I'll actually run automated tests?

oneroyalace:
@reefdog Rebased!

app/models/archive_item.rb
```diff
@@ -33,13 +33,33 @@ def self.create_from_media_review(media_review)
     # If so, skip the scrape, create the MediaReview and mark as `orphaned`.
     object_model.create_from_url(url) unless object_model.nil? # Start scraping
 
-    # For the moment we create an "orphan" MediaReview, since the archive_item will be created
-    # later. This will be updated with the associated Scrape gets fulfilled
-    MediaReview.create!(
+    # For the moment we create an "orphan" MediaReview without a parent ArchiveItem
+    # The ArchiveItem will be created after Zenodotus receives a callback from Hypatia
+    # That callback will triger the Scrape fulfillment processwhich
```
reefdog, line 38: `triger` → `trigger`

app/models/archive_item.rb
```diff
@@ -33,13 +33,33 @@ def self.create_from_media_review(media_review)
...
+    # which will atttach this MediaReview to the ArchiveItem
```
reefdog, line 39: `atttach` → `attach`

app/models/archive_item.rb
oneroyalace: Ty! I really need to add a spell checker to my text editor

app/models/archive_item.rb
oneroyalace: More letters better, right?

reefdog: Absollutelyy!

db/migrate/20220926191035_update_media_review.rb
```diff
@@ -0,0 +1,10 @@
+class UpdateMediaReview < ActiveRecord::Migration[7.0]
+  def change
+    change_table :media_reviews do |t|
+      t.datetime :date_published, null: false
+      t.text :url, null: false
+      t.jsonb :author, null: false
+      t.jsonb :item_reviewed, null: false
```
reefdog, line 7: Okay so per our conversation in Slack and after some more thinking, I think capturing these null constraints here only and _not_ on the model will bite us.

Can you at least update the model to validate the presence of any required attributes? Whether we remove the null constraints as well is a larger architectural/design question; I'm leaning to "yes, only keep null constraints where necessary for data integrity (primary keys, required foreign keys, etc.)" but open to yours and @cguess's thoughts.

reefdog (changes requested):
Two thoughts here:

1. These migrations will currently only run on a clean database. If we want to be able to run them on an existing database, we'll need to remove the `null` constraints on the new columns. I genuinely don't know if that matters, just raising it.
2. See my inline comment re: the null constraints at all. Happy to be over-ruled if you and @cguess disagree, but I'm thinking that capturing a required model attribute only in the db schema and not on the model is a pattern we'll end up regretting. I'm separately a little skeptical we need to have null constraints on required fields in the database at all, but open to thoughts! (If we go with "we don't", then I know there are lots of null constraints that could be lifted.)

db/migrate/20220926191035_update_media_review.rb
oneroyalace: Addressed!

oneroyalace:
Agreed @reefdog. I've removed non primary key + foreign key null constraints from both models

app/models/claim_review.rb
```diff
@@ -0,0 +1,3 @@
+class ClaimReview < ApplicationRecord
+  belongs_to :media_review, optional: true, class_name: "MediaReview"
```
reefdog, line 2: Do we want to add any model validations to `ClaimReview`? Not suggesting we should, just wondering.

app/models/archive_item.rb
```diff
@@ -33,13 +33,33 @@ def self.create_from_media_review(media_review)
     # If so, skip the scrape, create the MediaReview and mark as `orphaned`.
     object_model.create_from_url(url) unless object_model.nil? # Start scraping
 
-    # For the moment we create an "orphan" MediaReview, since the archive_item will be created
-    # later. This will be updated with the associated Scrape gets fulfilled
-    MediaReview.create!(
+    # For the moment we create an "orphan" MediaReview without a parent ArchiveItem
+    # The ArchiveItem will be created after Zenodotus receives a callback from Hypatia
+    # That callback will trigger the Scrape fulfillment processwhich
```
reefdog, line 38: `processwich` → `process which`

I am the 👑 of substantive suggestions.

db/migrate/20220926191035_update_media_review.rb
```diff
@@ -0,0 +1,15 @@
+class UpdateMediaReview < ActiveRecord::Migration[7.0]
```
reefdog, line 1: Not worth resolving here, but I'd aim for a more descriptive migration name in the future just for sanity. 😆

reefdog (approved):
This all looks great to me and the tests run!

I called out a comment typo, asked one question, and made one future suggestion, but none are crucial and this otherwise LGTM.

app/models/archive_item.rb
oneroyalace: Someday I'm actually going to install a spellchecker

reefdog: But then what will become of my code reviews??

Issue #301, opened 2022-08-05
Split site into both Insights and Vault


reefdog:
Our architecture calls for this single Rails app to serve two masters: Fact-Check Insights and MediaVault.

Insights will be the primary/default app served, available at `www.factcheckinsights.org`. MediaVault will be available at `vault.factcheckinsights.org` (but with a marketing-friendly redirect URL, e.g. `mediavault.beauty`). By sharing the same underlying Rails app, the bulk of the infrastructure can be shared. By sharing the same primary domain, the authentication session can be shared so users logged in to Vault are also logged in to Insights.

This task will entail:
- [x] Organizing and namespacing the app-specific controllers, views, tests, routes, etc.
- [x] Customizing the shared layouts/templates based on app
- [x] Fixing how ActionMailer generates URLs (so they're app-specific)
- [x] Tagging applicants with source app (so Vault applicants are noted as such)
- [x] Making sure Vault-sourced applicants receive Vault access
- [x] Limiting Vault access to Vault-permitted users
- [x] Enable linking back and forth

PR #374, opened 2022-09-28
Fix MediaVault routes to remove `media_vault` path

reefdog:
b7df27d3d8fb1f54f8a7b2f4e06241da81ba6dd8 intentionally namespaced the MV-related routes under `media_vault`, but this was meant to prefix the helpers (`media_vault_foo_bar_url`) and the controllers (`MediaVault::FooController`). It was _not_ intended to insert `media_vault/` into each path, which the `namespace` approach does.

This PR fixes that by using `scope` instead.

**Testing:**
- `rails t` (of course)
- Start the app, log in, and click "Text Search". The URL should just be `{hostname}/text_search`. If it is `{hostname}/media_vault/text_search`, then things are still screwy.

Issue #301

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -25,7 +25,7 @@ Rails.application.routes.draw do
     get "/confirm/done", to: "applicants#confirmed", as: "applicant_confirmed"
   end
 
-  namespace "media_vault" do
+  scope module: "media_vault", as: "media_vault" do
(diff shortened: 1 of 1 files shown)


reefdog:
I'm just gonna merge this one in, it's so unremarkable and really a hotfix.

PR #376, opened 2022-09-29
Serve Vault and Insights on discrete domains

reefdog:
🚨 **This is a PR with big consequences!**

Prior to this PR, things were nice and easy, as the entire app was hosted at a single domain. Locally, you could just fire up `http://localhost:3000` and be merry. Given our desired architecture, though, we want the app to appear as "Fact-Check Insights" when accessed via `www.factcheckinsights.org` and as "MediaVault" when accessed via `vault.factcheckinsights.org`. (They're both rooted at the same domain for session-sharing ease.)

Starting with this PR, we split the entire experience based on which subdomain you're using. While we fall back to the Insights experience (so `localhost` technically still works), if you want to see the MediaVault stuff — which is currently most of the site — you'll need to change your local setup.

- **To run Zenodotus**, you'll have to point `www.factcheckinsights.local` and `vault.factcheckinsights.local` to your localhost. I use `/etc/hosts`. If you want to get fancier with a local nginx and reverse proxy or whatnot, feel free.
- **To scrape / run Hypatia**, you'll also have to modify `URL` on the Zeno side to match (`http://vault.factcheckinsights.local:3000`), and modify `ZENODOTUS_URL` on the Hypatia side to match (`http://vault.factcheckinsights.local`). ⚠️ We should almost certainly rename our `URL` env var in Zeno to something more explicit, such as `MEDIA_VAULT_URL`.

After you've done all this, then when you fire up the app, you'll get to it from either `www.factcheckinsights.local` or `vault.factcheckinsights.local`, depending on which site you want. Logging in to either will log you in to the other as well, as the session is shared. Also, the Vault routes (e.g., `/archive`) are inaccessible at the `www` subdomain, by design.

⚠️ If your browser forces HTTPS, you'll want to make an exception for these domains.

**Testing:**
1. Do the prereq setup mentioned above to point the domains.
2. `rails t` of course
3. `rails s` and poke around the app at both domains, noting especially:
   - The domain-specific header and footer
   - The shared session between domains
   - The inaccessible Vault paths when on the Insights domain
4. Archive a URL to make sure scraping still works.

**Still to come:**
- The mailer uses the `URL` env var to construct URLs, which is currently pinned to Vault. This means all applications will get URLs that are rooted at `vault.factcheckinsights.local`. That will change so the mailer URLs are context-specific.
- Currently, all applicants look the same, but I will actually be storing which subdomain the applicant came from so the admin knows what site they're requesting access _to_. (Admins will be able to approve Vault access for any applicant even if they came from Insights, but this will be important context.)
- Actual content pages; right now the Insights nav is empty (aside from user stuff) and the Vault one is incomplete.

Issue #301

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,9 +133,14 @@ A faster image manipulation library than ImageMagick.
    1. For local development, `touch config/application.yml` and ask another developer for the config values
    1. For production, make sure the environment variables are set properly
 1. For local development, bootstrap the assets: `rails assets:precompile`
+1. Add the following entries to your `/etc/hosts` file:
+   ```
+   127.0.0.1	www.factcheckinsights.local
+   127.0.0.1	vault.factcheckinsights.local
+   ```
 1. In your shell, run `rails s` (to fire up just the Puma server) or `./bin/dev` (to also fire up the JS/CSS bundler, if you'll be modifying those assets)
 
-✨ The app should now be running and available at [http://localhost:3000](http://localhost:3000). If not, contact @cguess or another developer.
+✨ The app should now be running and available at [http://www.factcheckinsights.local:3000](http://www.factcheckinsights.local:3000) (Insights) and [http://vault.factcheckinsights.local:3000](http://vault.factcheckinsights.local:3000) (MediaVault). If not, contact @cguess or another developer.
 
 #### Starting the scraper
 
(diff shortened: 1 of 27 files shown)


oneroyalace (approved):
LGTM. Everything works locally

reefdog:
I should note that I don't know what the rollout to staging will look like for this. AFAIK @cguess has already pointed `staging.www.factcheckinsights.org` and `staging.vault.factcheckinsights.org` to the Heroku app, but I'm not sure what else we need to do on the Heroku config side.

reefdog:
Received @cguess's approval to bypass his review on this one. LEEEEROY JENKINS

PR #385, opened 2022-10-06
Handle archive.org links

### LINKED ISSUE #380: Handle Archive.org links
A number of entries use archive.org link instead of twitter.com or whatever. This obviously breaks everything. We should look into extracting urls from archive.org urls or extracting it straight from their archive (way harder).

oneroyalace:
This PR gives Zenodotus the ability to archive media re-hosted on archive.org 

If Zenodotus receives an Archive request for a media post hosted on archive.org, it will now try to extract the post's original URL and use it to generate a Hypatia scrape request. 

# Testing instructions
`rake`

Closes #380

### DIFF
--- a/test/media_sources/media_source_test.rb
+++ b/test/media_sources/media_source_test.rb
@@ -35,4 +35,10 @@ class MediaSourceTest < ActiveSupport::TestCase
   test "Can archive media using archive.org URLs" do
     assert_not_nil Sources::Tweet.create_from_url!("https://web.archive.org/web/20220930151009/https://twitter.com/AmtrakNECAlerts/status/1397922363551870990")
   end
+
+  test "archive.org URL extraction does not affect normal URLs" do
+    tweet_url = "https://twitter.com/AmtrakNECAlerts/status/1397922363551870990"
+
+    assert_equal tweet_url, MediaSource.extract_post_url_if_needed(tweet_url)
+  end
 end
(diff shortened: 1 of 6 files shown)


test/media_sources/media_source_test.rb
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    archive_to_direct_url_map.each  do |archive_url, post_url|
+      assert_equal post_url, MediaSource.extract_post_url_if_needed(archive_url)
+    end
+  end
+
+  test "Can archive media using archive.org URLs" do
+    assert_not_nil Sources::Tweet.create_from_url!("https://web.archive.org/web/20220930151009/https://twitter.com/AmtrakNECAlerts/status/1397922363551870990")
+  end
```
reefdog, line 37: We should probably add one more test to make sure `extract_post_url_if_needed` really does return non-archive.org tests untouched.

reefdog (approved):
Looks good and tests run clean!

I added one note about a nice-to-have test — ensuring the new method does return non-Archive URLs untouched. (Right now that behavior is implicitly tested through all the other source tests, but it'd be good to do it explicitly.) But wanted to just suggest that as an optional addition that doesn't require re-review.

test/media_sources/media_source_test.rb
oneroyalace: Great suggestion. I'll add that in!

Issue #290, opened 2022-08-04
Remove Node from the asset pipeline (again)


reefdog:
Quick backstory: Rails 7's embrace of importmaps makes a Node-free asset pipeline possible. We initially embraced it (#187, #188) before reverting back (#191) because I thought we'd use Flowbite. But after looking at the designs, I think we won't use it¹, which puts us back on the path to not needing Node.

This task, to be done while doing #267, is for recreating the work we did to replace jsbundling/cssbundling with importmaps and reduce our Node usage down to just a dev dependency (for ESLint).

¹The overhead of using them as HTML elements (rather than React components) is too high to justify them, especially when our needs are so relatively low.

Issue #267, opened 2022-07-28
Add final content and design


reefdog:
With the stubs in place (#301) and the designs (#320) and content delivered, apply the styling and content.

PR #393, opened 2022-10-07
Implement designed navbar

reefdog:
This PR slow-rolls our way into the new design by way of the navbar, which goes from this:
header-old

To this:
header-new

**Testing:**
- 🚨 `rails assets:precompile`
   - Something about the change to "real" domains broke live asset delivery. I suspect I'll be able to fix it when I purge Node again as part of #290.
- `rails s`
- Make sure the navbar looks like the [design](https://xd.adobe.com/view/082065b6-ada4-454e-a03c-80defb22f2e7-5d73/) (within reason)
- Log in and note the "user" menu; switch between Insights and Vault in the user menu
- Narrow the window to trigger the mobile viewport

⚠️ Note that everything below the navbar (content, footer, etc.) is either unchanged or still in flight.

Affects #267


reefdog:
I have received the +1 from @cguess and @oneroyalace to captain the front-end ship with more selective PR review only at major visual/functional milestones, so self-approving this one.

PR #394, opened 2022-10-11
Use the native Rails 7 asset pipeline and importmaps

### LINKED ISSUE #290: Remove Node from the asset pipeline (again)

reefdog:
This has been a storied journey! After initially doing this very thing in #189, we reverted back in #192 after discovering what I thought was a crucial limitation: the inability to use Flowbite (a front-end component library written in Tailwind). Since getting the designs delivered and actually implementing them in #267, it's clear we don't need Flowbite, which tips the first domino toward no longer needing Node for front-end asset bundling, meaning we can fully embrace the Rails 7 native preference for import maps.

So, this PR:

- Replaces all our JS and CSS bundling with the native Rails 7 method (asset pipeline + import maps).
- Reduces Tailwind down to just a convenience library of utility classes, rather than also serving as our CSS bundler. Switches from the Node package to the Ruby gem.
- Reduces our entire use of Node to just a dev dependency for linting purposes through ESLint.
   - Arguably, we could instruct developers to use global ESLint and remove Node from this project entirely, but that seems like a needless optimization. Doing that would mean we were still _requiring_ it, but trading the simplicity of `yarn install` for "RTFM". I'm open to that conversation but we can have it later.
- Fixes a couple of ESLint-discovered JS linting errors.
- Removes some unused boilerplate or vestigial files.

**Setup/Prerequisites:**
- Clean out some old things (not entirely necessary, but good for the soul and avoiding the occasional red herring):
   - Delete `node_modules/`
   - Delete `public/assets/`
   - Delete the **visible contents** of `app/assets/builds/` (but leave the directory and its `.keep` file)
- `bundle install`
- `yarn install`
- `rails tailwindcss:build` — I'm not using Tailwind myself, but we're still including it since the existing unconverted interface still uses it.

**Testing:**
- `rails t` of course
- `rails s` and poke around, making sure styles and JavaScript work (like the hamburger menu in the mobile navbar)
- To make sure style changes are available immediately, make a change to `app/assets/stylesheets/application.css`, such as adding `body { background: red; }`, saving, and reloading the site (**without** restarting the server). If you see the change, ✅ great. If you don't see the change, ❌ not great.
   - ⚠️ Changes to `application.tailwind.css` are only compiled if you're running `./bin/dev` rather than `rails s`, since only the former fires up the Tailwind watcher. You don't need to test this, it's just an FYI.

Closes #290


oneroyalace (approved):
LGTM. I don't understand the controller eager-loading stuff, but that's just because I don't know anything about front-end development :)). 

By the way, I had to run `rails assets:precompile` in order to get the site to display properly. Should that be necessary? I had that experience on two different machines.

reefdog:
> I don't understand the controller eager-loading stuff, but that's just because I don't know anything about front-end development :)).

lol, well that part's just some Rails (well, Stimulus gem) boilerplate for preemptively importing all the Stimulus controllers in that directory. An alternative to it would be running a rake task that scanned the directory and generated explicit import statements, but I checked what a new Rails app did and it used the eager-loading. No big deal either way.

> By the way, I had to run `rails assets:precompile` in order to get the site to display properly. Should that be necessary? I had that experience on two different machines.

Huh. No, that shouldn't have been necessary. Had you run the `rails tailwindcss:build` task that I added to the description late?

reefdog:
Whoops, turns out my original PR description had a typo, apologies @oneroyalace!

The correct directory to delete is `public/assets`, **not** `app/public/assets` (which doesn't exist). Heads up @cguess (and I've updated the PR desc).

PR #399, opened 2022-10-17
Mark scrapes as completed but removed if so

### LINKED ISSUE #398: Add ability to receive "post removed" callback from Hypatia
We need to add the ability to kill a scrape request instead of just resubmitting. This is following on https://github.com/TechAndCheck/hypatia/pull/72.

We'll mark a media review as "removed" when received and make sure it's not resubmitted. Also, probably good to put a failsafe to disallow a scrape if "removed" is true just incase we mass resubmit at some point.

cguess:
This marks a scrape as fulfilled but removed if that's the callback from Hypatia. This then makes sure the scrape is not retried over and over and the MediaReview object also gets tagged as `taken_down`.

Note:
This adds a new migration
This adds a MediaReview test

To test:
- `rails db:migrate`
- `rails t`

closes #398


app/models/scrape.rb
```diff
@@ -24,23 +24,32 @@ def send_notification
...
-      self.update!({ fulfilled: true, archive_item: archive_item.first })
+    begin
+      # Mark removed if we're informed it no longer exists
+      if response.empty? == false && response.first.key?("status") && response.first["status"] == "removed"
+        self.update!({ fulfilled: true, removed: true })
+        media_review_item.update!({ taken_down: true })
+        return
+      end
+    rescue StandardError
+      # debugger
```
reefdog, line 40: Do we need to bail out early, or do anything else?

app/models/scrape.rb
```diff
@@ -24,23 +24,32 @@ def send_notification
...
     end
+
+    archive_item = ArchiveItem.model_for_url(self.url).create_from_hash(response)
+
+    # Update the previously created MediaReview item with the now archived stuff
+    # if archive_item.empty?
+    #   media_review_item.update!({ taken_down: true })
+    #   self.update!({ fulfilled: true })
+    # else
```
reefdog, line 49: Do we need to keep this commented-out stuff around? It looks like we sort of handle this branch up in the block above?

app/models/scrape.rb
cguess: nope, left in by accident.

reefdog (approved):
All seems good to me!

PR #412, opened 2022-10-25
Send a notification when a scrape is done

### LINKED ISSUE #410: Admin page not being updated when a scrape is fulfilled
The total count of scrapes still out in the world is not updated as scrapes are fulfilled. The table should be updated as well.

cguess:
We somehow forgot to update the `scrapes` channel when a scrape is fulfilled. This fixes that.

To test:
Unfortunately the only way to test this is to launch a Hypatia instance and go through the whole thing of scraping and waiting for a callback. The code that's called is already in production use, so maybe we just say "ok" to it ;-)`

Closes #410


reefdog (approved):
I can't imagine this is not a good thing. :D

Issue #391, opened 2022-10-07
Generate MediaReview schema output


reefdog:
We need to be able to generate a MediaReview representation that matches this output:

```json
{
	"@context": "https://schema.org/",
	"@type": "MediaReview",
	"datePublished": "2021-04-27",
	"url": "https://www.politifact.com/factchecks/2021/apr/27/instagram-posts/mariah-carey-didnt-fake-getting-her-covid-19-vacci/",
	"author": {
		"@type": "Organization",
		"name": "PolitiFact",
		"url": "https://politifact.com/"
	},
	"mediaAuthenticityCategory": "DecontexualizedContent",
	"originalMediaContextDescription": "Singer Mariah Carey shared a video of herself receiving a COVID-19 vaccination.",
	"itemReviewed": {
		"@type": "MediaReviewItem",
		"creator": {
			"@type": "Person",
			"name": "Instagram user",
			"url": "https://www.instagram.com/wrong_saloon_bear/?hl=en"
		},
		"interpretedAsClaim": {
			"@type": "Claim",
			"description": "Mariah Carey faked getting her COVID-19 vaccine because the needle can't be seen coming out of her arm."
		},
		"mediaItemAppearance": [
			{
				"@type": "VideoObjectSnapshot",
				"description": "An Instagram user posted a zoomed-in version of a video of Mariah Carey receiving a COVID vaccination, writing 'It's all a scam, don't celebrate celebrities.'"
			},
			{
				"@type": "VideoObjectSnapshot",
				"contentUrl": "https://www.instagram.com/p/CNtao_WA0Dr/",
				"archivedAt": "https://archive.is/EXAMPLE"
			}
		]
	}
}
```

oneroyalace:
Closed by #407

Issue #390, opened 2022-10-07
Generate ClaimReview schema output


reefdog:
We need to be able to generate a ClaimReview representation that matches this output:

```json
{
	"@context": "https://schema.org/",
	"@type": "ClaimReview",
	"datePublished": "2014-07-23",
	"url": "http://www.politifact.com/texas/statements/2014/jul/23/rick-perry/rick-perry-claim-about-3000-homicides-illegal-immi/",
	"author": {
		"@type": "Organization",
		"url": "http://www.politifact.com/",
		"image": "http://static.politifact.com/mediapage/jpgs/politifact-logo-big.jpg",
		"sameAs": "https://twitter.com/politifact"
	},
	"claimReviewed": "More than 3,000 homicides were committed by \"illegal aliens\" over the past six years.",
	"reviewRating": {
		"@type": "Rating",
		"ratingValue": 1,
		"bestRating": 6,
		"image": "http://static.politifact.com.s3.amazonaws.com/rulings/tom-pantsonfire.gif",
		"alternateName": "True"
	},
	"itemReviewed": {
		"@type": "CreativeWork",
		"author": {
			"@type": "Person",
			"name": "Rich Perry",
			"jobTitle": "Former Governor of Texas",
			"image": "https://upload.wikimedia.org/wikipedia/commons/thumb/1/15/Gov._Perry_CPAC_February_2015.jpg/440px-Gov._Perry_CPAC_February_2015.jpg",
			"sameAs": [
				"https://en.wikipedia.org/wiki/Rick_Perry",
				"https://rickperry.org/"
			]
		},
		"datePublished": "2014-07-17",
		"name": "The St. Petersburg Times interview [...]"
	}
}
```

ericaryan:
When possible, can I get an excerpt of the download with multiple examples so I can double-check the schema table fields are labeled/explained correctly? I wanted to wait to compare it with the "real" output before calling it done.

reefdog:
@ericaryan Will do! The full feed is coming in PR #417 as part of issue #275, so we should be able to get you a sample very soon.

PR #417, opened 2022-10-27
275 insights exports

### LINKED ISSUE #275: Build authorized data download
Authenticated users should be able to download the Fact-Check Insights data.

- [x] Determine exactly what content should be available in the download @reefdog 
- [x] Write the query that generates that data @cguess @oneroyalace 
- [x] Make sure the query is cached by Rails @cguess @oneroyalace 
- [x] Create the protected endpoint for grabbing that data @reefdog 
- [x] Add the download button to the Download page @reefdog 

We expect that the data will not be complete at first, as we will have to continue building out the scraping infrastructure. This issue is complete when the user can generate the download, not necessarily when 100% of the data is available.

oneroyalace:
This PR adds `MediaReview`/`ClaimReview` data export functionality to Zenodotus. It uses `BluePrinters` specs created per #390 and #391 to generate JSON exports and a new gem, `comma`, to generate CSV exports. 

The CSV exports are a bit hairy, because they're flat representations of nested data. Column names in the CSV use use underscores to denote nested attributes. 

For example, data corresponding to the `reviewRating.ratingValue` attribute will be written to a  column named `reviewRating_ratingValue`. 

Closes #275 

# Testing instructions
- `bundle` (install `comma`)
- `rake`
- Log into FCI and try out the CSV and JSON data downloads. Make sure the data in the downloaded files matches the data in our `seeds.rb` file.


app/controllers/fact_check_insights_controller.rb
```diff
@@ -17,22 +17,17 @@ def download
...
-        )
+        send_data(generate_json, type: "application/json", filename: "fact_check_insights.json")
       end
+
       format.zip do
         unless user_signed_in? && current_user.can_access_fact_check_insights?
-          render "You do not have permission to view that resource.", status: :unauthorized
+          render json: {
+            error: "You do not have permission to view that resource."
+          }, status: :unauthorized
```
reefdog, line 27: This should render a plaintext response (but still with the unauthorized status) rather than JSON, I think? Since the client didn't request JSON.

reefdog (changes requested):
Okay, so this is largely great work. I'm running into a few issues:

1. While tests run clean, actually trying to download the data through the browser throws `undefined local variable or method 'generate_csv_zip'` or `undefined local variable or method 'generate_json'`. So the problem there is two-part: that the buttons aren't downloading, and that the tests didn't catch it. (Unless they are, and it's my system that's broken! Can you verify on your end?)
2. The commit history on this branch seems to have gotten garbled from a merge or rebase. Your relevant commits are very well-organized, so I'd love for the final merge to be just as clean. My method for this (tested locally) would be to make a new branch from `master`, and then selectively cherry-pick your commits onto it, and then force-push to this origin branch. (Or, worst case scenario, just open a new PR with that new/clean branch.) My Git GUI (Tower) makes this super easy, but I don't know how difficult it is on the CLI or other clients. Feel free to ring me to pair on this if you'd like!
3. One minor issue noted inline: it'd be good for the CSV download action to return a non-JSON response for the unauthorized case.

PR #426, opened 2022-11-01
Fact Check Insights exports

### LINKED ISSUE #275: Build authorized data download

oneroyalace:
This PR adds `MediaReview`/`ClaimReview` data export functionality to Zenodotus. It uses `BluePrinters` specs created per #390 and #391 to generate JSON exports and a new gem, `comma`, to generate CSV exports. 

The CSV exports are a bit hairy, because they're flat representations of nested data. Column names in the CSV use use underscores to denote nested attributes. 

For example, data corresponding to the `reviewRating.ratingValue` attribute will be written to a  column named `reviewRating_ratingValue`. 

Closes #275 

# Testing instructions
- `bundle` (install `comma`)
- `rake`
- Log into FCI and try out the CSV and JSON data downloads. Make sure the data in the downloaded files matches the data in our `seeds.rb` file. Try loading the download URLs without being logged in. You shouldn't get any data :)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -183,3 +183,6 @@ gem "tailwindcss-rails", "~> 2.0"
 
 # We use SCSS to write more CSS more concisely
 gem "dartsass-rails", "~> 0.4.0"
+
+# Comma lets us generate CSV-formatted data from ActiveRecord models
+gem "comma", "~>4.7.0"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -137,6 +137,8 @@ GEM
     childprocess (4.1.0)
     climate_control (0.2.0)
     coderay (1.1.3)
+    comma (4.7.0)
+      activesupport (>= 4.2.0)
     commander (4.6.0)
       highline (~> 2.0.0)
     concurrent-ruby (1.1.10)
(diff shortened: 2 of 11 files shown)


reefdog (approved):
This all looks great to me. I didn't do a deep dive on data fidelity; I think we can let @joelwluther and @ericaryan do that as part of the QA once this is on staging (i.e. has a wider sampling of data). But overall, great.

Can you pop over to #417 and delete the old branch? I would, but want to let you confirm it's really and truly okay to do so.

reefdog:
@oneroyalace Oh, and after playing around with the idea for rendering the ZIP error in `plain`, I think you're right to leave it as a JSON response. The browser doesn't seem to do what I expect when sending a plain response with an unauthorized status. It's all fine.

PR #427, opened 2022-11-02
Update `MediaReview` CSV serializer to handle null attribute

oneroyalace:
This PR handles cases in which the `MediaReview.itemReviewd.mediaItemAppearances` attribute is `nil` or an empty array by just returning `nil` for all related attributes in the CSV

# Testing
`rake`

### DIFF
--- a/app/models/media_review.rb
+++ b/app/models/media_review.rb
@@ -38,19 +38,32 @@ class MediaReview < ApplicationRecord
 
     # The CSV representation of a MediaReview will only include a single object from the mediaItemAppearance array
     # Namely, the object containing the `contentUrl` attribute
+    # If the mediaItemAppearance array is nil or empty, we'll return null values for all related attributes in the CSV
     item_reviewed "itemReviewed_mediaItemAppearance_@type" do |item_reviewed|
-      appearance = item_reviewed.dig("mediaItemAppearance").filter { |appearance| appearance.has_key?("contentUrl") }.first
-      appearance["@type"]
+      if !item_reviewed.has_key?("mediaItemAppearance") || item_reviewed["mediaItemAppearance"].empty?
+        nil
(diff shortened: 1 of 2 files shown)


app/models/media_review.rb
```diff
@@ -38,19 +38,32 @@ class MediaReview < ApplicationRecord
...
+    # If the mediaItemAppearance array is nil or empty, we'll return null values for all related attributes in the CSV
     item_reviewed "itemReviewed_mediaItemAppearance_@type" do |item_reviewed|
-      appearance = item_reviewed.dig("mediaItemAppearance").filter { |appearance| appearance.has_key?("contentUrl") }.first
-      appearance["@type"]
+      if !item_reviewed.has_key?("mediaItemAppearance") || item_reviewed["mediaItemAppearance"].empty?
+        nil
+      else
+        appearance = item_reviewed.fetch("mediaItemAppearance", []).filter { |appearance| appearance.has_key?("contentUrl") }.first
+        appearance["@type"]
+      end
```
reefdog, line 48: This is a good guard! However, I think the more idiomatic way to write it would be just to return early rather than do a full if/else block, ala:

```ruby
return nil if !item_reviewed.has_key?("mediaItemAppearance") || item_reviewed["mediaItemAppearance"].empty?

appearance = item_reviewed.fetch("mediaItemAppearance", []).filter { |appearance| appearance.has_key?("contentUrl") }.first
appearance["@type"]
```

If it's not much trouble to edit this commit or just patch on top of it, can we do so?

reefdog: Scratch this, @oneroyalace has pointed out it… doesn't work! 🤦

reefdog (approved):
This tests clean and resolves a real issue. 👍

PR #430, opened 2022-11-03
Create `AdminController` and put JobsStatus stuff inside

reefdog:
In preparation for finishing #273, we need to formally have an admin area. This PR adds the `AdminController` and (very minimal) dashboard, and moves the `JobsStatusController` inside. It also fixes a few unrelated but discovered-while-doing-it bugs.

**Testing:**
- `rails t`
- Run the app and try to access `/admin` while not logged in or while logged in as a non-admin. ❌ It shouldn't work.
- Now go there while logged in as an admin. ✅ It should work.
- Make sure the "Admin" item in the user menu (only visible when logged in) goes to the new Admin "dashboard" rather than directly to the Jobs Status page the way it used to.
- Make sure you can still use the Jobs Status dashboard as before (with no UX regressions from `master`).

Affects #273


oneroyalace (approved):
Sorry for the late review. All LGTM

Issue #364, opened 2022-09-23
Improve search and browse design


reefdog:
This is an umbrella issue for tracking a constellation of improvements to the search and browse design.

- [x] #438
- [x] #440 
- [x] #439
- [x] #472
- [x] #473 
- [x] #365

PR #445, opened 2022-11-07
Add MediaVault text search to navbar

### LINKED ISSUE #438: Add text search to navbar

reefdog:
This PR adds the text search functionality directly to the navbar:

MediaVault navbar

It also removes the dedicated search page (although the search _results_ page remains), as searching will be spawned from the navbar and from a more advanced search atop the Dashboard page.

This PR also sneaks in a move of the Add URL button to the user menu to make more room on the primary navbar.

**Testing:**
- `rails t`
- Start the server with `./bin/dev` (or `rails assets:precompile` + `rails s`)
- Log in and search for things
- ✅ Get search results as expected
- ✅ See query prefilled in the search field on the search results page
- ✅ See search history still works (in user menu > Settings)
- ✅ Narrow the window enough to spawn the mobile menu and ensure the search form shows up there as expected

⚠️ Note that media search is not affected and the search results screen is not yet changed/improved.

Closes #438


cguess:
Two things as I begin the review: can we change "Media Search" to "Reverse Media Search"? It'll make the functionality much clearer without onboarding. Second: Not now, but let's consider how to add media search functionality into the search bar (see: google images) so that there's not two distinct search pages.

cguess:
This is what I'm getting on Safari and Firefox and Chrome

Screenshot 2022-11-08 at 8 38 22 AM

reefdog:
> Can we change "Media Search" to "Reverse Media Search"? It'll make the functionality much clearer without onboarding.

Sure! I'll default to that language but may explore even more clear language. I'll do that as another sub-task of #364 rather than as part of this PR though.

> Not now, but let's consider how to add media search functionality into the search bar (see: google images) so that there's not two distinct search pages.

Absolutely. I have some ideas along those lines as part of #439, where we have a single UI element that allows easy toggling back and forth.

> This is what I'm getting on Safari and Firefox and Chrome…

Hmm. Did you happen to start the server with `rails s`? I think you'll either need to start with `./bin/dev` or do the `rails assets:precompile` first. I keep forgetting to note that. It's a bummer and I wish we didn't need to do that. :/

Can you give that a try?

cguess (approved):
See, this is the type of commit that would get you fired from Twitter. Can't go around deleting useless code like that!

Anyways, it's confirmed aside from the gray border issue.

reefdog:
Ooh, nice catch on the gray border. No, I think I just had a blind spot to it. 😬 I'll make that styling change. Thanks for the review!

PR #447, opened 2022-11-08
Rename ingest endpoint

### LINKED ISSUE #446: Rename ingest method
We're adjusting the ingest endpoint to accept media review and claim review at the same time. So we'll rename it.

cguess:
This renames the ingestion endpoint and allow ClaimReview or MediaReview to be submitted. It adds schema validation for ClaimReview and also removes a `media_review_id` null constraint that was on the ClaimReview model. Otherwise there should be no changed functionality.

To test:
`rails db:migrate`
`rails t`

closes  #446

### DIFF
--- a/app/controllers/media_vault/ingest_controller.rb
+++ b/app/controllers/media_vault/ingest_controller.rb
@@ -79,12 +79,12 @@ class MediaVault::IngestController < MediaVaultController
   end
 
   # A class representing the allowed params into the `submit_media_review` endpoint
-  class SubmitMediaReviewParams < T::Struct
-    const :media_review_json, String
+  class SubmitReviewParams < T::Struct
+    const :review_json, String
   end
 
   # A class representing the allowed params into the `submit_media_review_source` endpoint
-  class SubmitMediaReviewSourceParams < T::Struct
+  class SubmitReviewSourceParams < T::Struct
     const :url, String
   end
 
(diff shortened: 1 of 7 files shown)


app/models/claim_review.rb
```diff
@@ -1,5 +1,5 @@
 class ClaimReview < ApplicationRecord
-  validates :media_review_id, presence: true
+  # validates :media_review_id, presence: true
```
reefdog, line 2: Can we just… delete this?

db/migrate/20221108194425_remove_null_contraint_from_claim_review_for_media_review.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveNullContraintFromClaimReviewForMediaReview < ActiveRecord::Migration[7.0]
```
reefdog, line 1: Small thing but typo in this migration name (`contraint`)

reefdog (changes requested):
Largely this is fantastic. I added two inline comments that are non-blocking (thus not attached to this review). The only requested change is that one test is failing, but I think it's because we should actually remove this test entirely now:

```
Failure:
ClaimReviewTest#test_can't_create_claim_review_without_mediareview [/Users/justin/Projects/duke/zenodotus/test/models/claim_review_test.rb:81]:
ActiveRecord::RecordInvalid expected but nothing was raised.

rails test test/models/claim_review_test.rb:80
```

Once that test is removed/resolved, feel free to merge in at your leisure!

app/models/claim_review.rb
cguess: yes, damn I missed it (had to juggle a lot between here and FactStream)

app/controllers/media_vault/ingest_controller.rb
```diff
@@ -224,4 +263,17 @@ def validate_media_review(media_review)
   rescue StandardError
     false
   end
+
+  # Validate MediaReview that was passed in
```
reefdog, line 267: Oops sorry, mind correcting this comment? It says MediaReview when it's for ClaimReview.

Or, I mean, delete it. It's not really necessary, the method is clearly named.

PR #451, opened 2022-11-10
Enable updates to `MediaReview`/`ClaimReview`

### LINKED ISSUE #449: Update `ClaimReview`/`MediaReview` records when they're re-submitted to Zenodotus
If a fact checker revises their check, we want to update our `ClaimReview`/`MediaReview` records. @cguess is updating factstream to send over a `factstream_id` attribute, which we'll store in our `ClaimReview`/`MediaReview` models as `external_unique_id`. When we receive a factstream object with an `external_unique_id` already in our DB, we'll update the matching record with the new data. 

to-dos here:
1. add an indexed `external_unique_id` attribute to the `ClaimReview` and `MediaReview` models
2. Add logic to update records when we ingest an object from factstream that has a previously recorded `external_unique_id`

oneroyalace:
In tandem with [FactStream PR] this PR lets us update `ClaimReview` and `MediaReview` records when we receive revisions to them from `FactStream`. 

We recognize those revisions by storing the new `external_unique_id` attribute `FactStream` attaches to `ClaimReview` and `MediaReview` and using it to see whether we've already stored identically id'd objects. 

# Testing instructions
1. `rails db migrate` (add indexed `external_unique_id` column to `ClaimReview` and `MediaReview`)
2. `rake`

Closes #449

### DIFF
--- a/app/controllers/media_vault/ingest_controller.rb
+++ b/app/controllers/media_vault/ingest_controller.rb
@@ -10,13 +10,15 @@ class MediaVault::IngestController < MediaVaultController
     extend T::Sig
 
     enums do
-      Success = new
+      Created = new
+      Updated = new
     end
 
     sig { returns(Integer) }
     def code
       case self
-      when Success then 20
+      when Created then 20
+      when Updated then 20
       else T.absurd(self)
       end
     end
@@ -24,7 +26,8 @@ class MediaVault::IngestController < MediaVaultController
     sig { returns(String) }
     def message
       case self
-      when Success then "Successfully archived media object"
+      when Created then "Successfully archived media object"
+      when Updated then "Successfully updated mediareview for existing object"
       else T.absurd(self)
       end
     end
@@ -81,6 +84,7 @@ class MediaVault::IngestController < MediaVaultController
   # A class representing the allowed params into the `submit_media_review` endpoint
   class SubmitReviewParams < T::Struct
     const :review_json, String
+    const :external_unique_id, String
   end
 
   # A class representing the allowed params into the `submit_media_review_source` endpoint
@@ -89,22 +93,25 @@ class MediaVault::IngestController < MediaVaultController
   end
 
 
-  # Creates a MediaReview and ArchiveItem based on a POSTed MediaReview JSON object
-  #
-  #  @params {media_review_json} A MediaReview JSON object
+  # Uses a JSON object to create a combination of ArchiveItem, MediaReview, and ClaimReview records
+  #  @params {review_json} A JSON object
+  #  @params {external_unique_id} A UUID
   sig { void }
   def submit_review
     # TODO: Spin off an active job to handle this
     typed_params = TypedParams[SubmitReviewParams].new.extract!(params)
     review_json = JSON.parse(typed_params.review_json)
+    external_unique_id = typed_params.external_unique_id
 
     raise ApiErrors::JSONValidationError.error unless review_json.key?("@type")
 
     case review_json["@type"]
     when "ClaimReview"
-      response_payload = archive_from_claim_review(review_json)
+      should_update = ClaimReview.where(external_unique_id: external_unique_id).present?
+      response_payload = archive_from_claim_review(review_json, external_unique_id, should_update)
     when "MediaReview"
-      response_payload = archive_from_media_review(review_json)
+      should_update = MediaReview.where(external_unique_id: external_unique_id).present?
+      response_payload = archive_from_media_review(review_json, external_unique_id, should_update)
     else
       raise ApiErrors::JSONValidationException
     end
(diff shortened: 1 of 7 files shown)


app/controllers/media_vault/ingest_controller.rb
```diff
@@ -208,48 +215,55 @@ def find_media_review_in_page(url)
...
   # @param [Hash] media_review_json: A MediaReview hash object
+  # @param [Boolean] should_update: A boolean indicating whether we should create or update a record
   # @return [Hash]: A hash containing response codes and a reference to the newly created ArchiveItem
-  sig { params(media_review_json: Hash).returns(Hash) }
-  def archive_from_media_review(media_review_json)
+  sig { params(media_review_json: Hash,
+               external_unique_id: T.nilable(String),
+               should_update: T::Boolean
+              ).returns(Hash) }
+  def archive_from_media_review(media_review_json, external_unique_id, should_update = false)
```
reefdog, line 224: What would you think about:

1. Converting `external_unique_id` and `should_update` to keyword arguments instead of positional, so they'll be more legible at invocation? And…
2. …defaulting `external_unique_id` to `nil`? That seems like a reasonable default, and would prevent having to specify every time when sending a new one.

I recognize both changes would have to be implemented across like half a dozen methods. I truly leave it up to you to decide if you want to make the change and if you think it's worthwhile!

app/models/media_review.rb
```diff
@@ -67,6 +67,51 @@ class MediaReview < ApplicationRecord
...
+    end.first
+
+    appearance["contentUrl"]
+  end
+
+  # Create or update a MediaReview Record using the input hash
+  sig { params(media_review_hash: Hash,
+               external_unique_id: T.nilable(String), # We won't have a uuid when we scrape mediareview from pages
+               should_update: T::Boolean
+              ).returns(MediaReview) }
```
reefdog, line 84: FWIW I prefer this multi-line sig definition syntax to the single-line version you use elsewhere. Shorter / more readable lines, plus easier diffs in future. 👍

app/models/media_review.rb
```diff
@@ -67,6 +67,51 @@ class MediaReview < ApplicationRecord
...
+        original_media_link: url,
+        media_authenticity_category: media_review_hash["mediaAuthenticityCategory"],
+        original_media_context_description: media_review_hash["originalMediaContextDescription"],
+        date_published: media_review_hash["datePublished"],
+        url: media_review_hash["url"],
+        author: media_review_hash["author"],
+        item_reviewed: media_review_hash["itemReviewed"]
+      )
+      existing_media_review.reload
+      existing_media_review
```
reefdog, line 100: I thiiiiink you can just end the block on the `.update!` call, right? Doesn't it return the (already reloaded) object? If so might save us a query.

app/models/claim_review.rb
```diff
@@ -31,6 +31,32 @@ class ClaimReview < ApplicationRecord
     item_reviewed "itemReviewed_author_sameAs" do |item_reviewed| item_reviewed.dig("author", "sameAs") end
   end
 
+  sig { params(claim_review_hash: Hash, external_unique_id: T.nilable(String), should_update: T::Boolean).returns(ClaimReview) }
+  def self.create_or_update_from_claim_review_hash(claim_review_hash, external_unique_id, should_update)
+    if should_update
+      existing_claim_review = ClaimReview.where(external_unique_id: external_unique_id).first
```
reefdog, line 37: Pretty sure `where(external_unique_id: external_unique_id).first` can be reduced to `find_by(external_unique_id: external_unique_id)`. I assume we're unique-indexing on this column and so will only get one thing back anyway, but… there you go. (I see `where().first` a few more times so in general this optimization may actually apply elsewhere.)

reefdog (changes requested):
Tests pass and I assume that means it will do what it says! 👍

I have a smattering of syntax notes and suggestions left inline and so am marking this "request changes". But none are critical, and so if you want to merge this in anyway, I humbly defer.

app/models/media_review.rb
```diff
@@ -67,6 +67,51 @@ class MediaReview < ApplicationRecord
     end
   end
 
+  # Extract the content URL from a MediaReview hash
+  sig { params(media_review_hash: Hash).returns(String) }
+  def self.get_content_url(media_review_hash)
+    appearance = media_review_hash["itemReviewed"]["mediaItemAppearance"].select do |appearance|
+      appearance.key?("contentUrl")
+    end.first
```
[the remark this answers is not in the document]
oneroyalace: Aha! Nice! I think I've used that in `Forki` before

app/models/media_review.rb
oneroyalace: Good point about the diffs! I'll change these

app/models/media_review.rb
reefdog: Oh interesting! I thought `update()` returned true or false and `update!()` returned the new object or threw an error. Clearly I was wrong, thanks for trying anyway!

PR #453, opened 2022-11-10
Update video search so it's quicker

cguess:
This updates Eikon allowing us to limit the number of searches for a video, while also reducing the nearness for results to count, since video frames will almost never line up perfectly.

To test:
`bundle install`
`rails t test/models/image_search.rb`


oneroyalace (approved):
LGTM

config/database.yml
```diff
@@ -84,4 +84,7 @@ test:
 # for a full overview on how database connection configuration can be specified.
 #
 production:
-  url: <%= ENV["DATABASE_URL"] %>
+  <<: *default
+  database: zenodotus_production
+
+  # url: <%= ENV["DATABASE_URL"] %>
```
reefdog, line 90: @cguess Whoops! Looks like this snuck past review, can you hotfix?

PR #454, opened 2022-11-11
Log user in after changing password

### LINKED ISSUE #355: Changing password logs out the user
There's a bug in the current password-changing code, unsure if it's old or has been hanging around a while, but:

1. Log in
2. Go to your user settings page
3. Change your password
4. Upon successful change, refresh the page

**What should happen:** You're still on the settings screen.
**❌ What does happen:** You get booted to the login screen.

Unsure how long this has been the case, but it predated the recent user roles stuff (just checked).

This should be addressed during #298.

cguess:
For some reason changing the password resets the users session and logs them out. Now we just programatically log them back in immediately.

To test:

On master branch:
1. Log in with a user
2. Change the password
3. Refresh the page - you'll end up at the log in page

Switch to this branch:
1. Log in with user
2. Change the password
3. Refresh the page - you should stay on the same page

closes #355

### DIFF
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -96,7 +96,11 @@ class AccountsController < ApplicationController
   sig { void }
   def change_password
     typed_params = TypedParams[ChangePasswordParams].new.extract!(params)
+
     current_user.reset_password(typed_params.password, typed_params.confirmed_password)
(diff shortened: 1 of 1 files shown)


oneroyalace (approved):
LGTM

reefdog (approved):
:pray:

Issue #456, opened 2022-11-12
Consider a better way to use Turbo for search


reefdog:
In a forthcoming commit as part of #440 (and based on the [design](https://github.com/TechAndCheck/zenodotus/issues/285)), we're moving the search into the page header. Consequently, using Turbo to fetch the results is a bit more difficult. The search results frames are only rendered on the search page itself, so we'd either need to:

1. Change how we use Turbo so that the entire main area of the page is a frame that gets refreshed (and then handle things like page title, nav state, etc.)
2. Conditionally enable Turbo when we're on a page that includes the necessary frames

① is more work than we can do right now. ② I thought would be easy, but I'm running into issues and have to stop.

But it's worth returning to the task at some point as an optimization. Meanwhile, storing the code (rather than leaving it in and commented-out).

In `SearchController#search_by_media` (`media_vault/search_controller.js`):
```ruby
respond_to do |format|
  format.turbo_stream do
    @media_results = @media_search.run

    # Add the media search ID so that we can adjust the URL and make the page reloadable
    response.headers["X-media-search-id"] = @media_search.id

    render turbo_stream: [
      turbo_stream.replace(
        "search_results",
        partial: "media_vault/image_search/results",
        locals: { search: @media_search, results: @media_results }
      ),
      turbo_stream.replace(
        "search_item",
        partial: "media_vault/image_search/search_item",
        locals: { search: @media_search, results: @media_results }
      )
    ]
  end
end
```

In any media search form:
```html
<form data-action="submit->media-vault--search-box#onMediaSearchSend turbo:submit-end->media-vault--search-box#onMediaSearchSuccess" />
```

In `search_box_controller.js`:
```javascript
export default class extends Controller {
  onMediaSearchSend(event) {
    if (this.searchByMediaFileInputTarget.files.length === 0) {
      alert('You need to choose an image or video to search by.')
      event.preventDefault()
    } else {
      this.searchByTextFieldsetTarget.disabled = true
    }
  }

  onMediaSearchSuccess(event) {
    // Reset the form so it can be used again.
    this.resetForms()

    // Update the URL and history with the media search ID
    const mediaSearchId = event.detail.fetchResponse.response.headers.get('X-media-search-id')

    const newUrl = new URL(window.location.href)
    const params = new URLSearchParams(newUrl.search)
    params.set('msid', mediaSearchId)
    newUrl.search = params
    window.history.pushState({}, null, newUrl.toString())
  }

  resetForms() {
[...]

PR #457, opened 2022-11-12
Combine text and image search into one controller and UX

### LINKED ISSUE #439: Allow drag-and-drop image search on text search field

### LINKED ISSUE #440: Combine text and media search into a single search controller
We currently have separate controllers for text and media searching. These should instead be under the same controller so that logic and templates can be shared better.

Along the way, let's rename anything about "image search" to be "media search" (unless it's _specific_ to searching by image instead of e.g. searching by video).

reefdog:
This rather large PR unifies the text and image search controllers into a single controller, and adds a search widget to the site header that handles both text and media searching:

https://user-images.githubusercontent.com/4731/201447389-a9ebcd0f-55e4-4ede-99af-02b52d64cf6d.mov

A few notes:
- **Search results are not updated** and still use the old templates (and in their old directory locations), this will come soon I promise but this PR is about getting the search widget experience working
- Users can search by media either by switching to the media mode by clicking the link, or by drag-dropping onto the widget (in either mode)
- I had to remove the Turbo frame search results for now (see #456), because we can't guarantee the frames will be on-page now that the widget is on the nav

**Testing:**
- `rails t`
- `rails assets:precompile` before `rails s`, _or_ start with `./bin/dev`
- Play with the search! Feel free to try and break it.

Closes #440
Closes #439


test/controllers/media_vault/search_controller_test.rb
```diff
@@ -0,0 +1,43 @@
...
+  test "creates text search history" do
+    sign_in users(:media_vault_user)
+
+    get media_vault_search_url(q: "Biden")
+
+    assert_response :success
+    assert_equal 1, TextSearch.count
+  end
+
+  # TODO: check that text search results are returned properly
```
oneroyalace, line 40: 👀 :

app/assets/stylesheets/search_box.scss
```diff
@@ -0,0 +1,103 @@
+/*
+Search Box
+
+A combination text field and submit button, that also allows drag-and-drop file uploading.
+
+```html
+<div class="search-box">
```
oneroyalace, line 7: Is this here for easy reference?

oneroyalace (approved):
LGTM. I noticed https://github.com/TechAndCheck/zenodotus/issues/459 while manually testing, but that's not connected to this PR

app/assets/stylesheets/search_box.scss
reefdog: Yeah, pretty much! If you poke through the other component SCSS files, you'll see that I try to show a representative block of markup (usually highly simplified, with no content or other attributes) so that anyone looking to understand the styles can quickly reference how they're expected to be structured.

I'm not sure that the stylesheet is the best place to document this, but honestly it's useful just for my own reference at times, to be able to glance up and remember "oh yeah, X is meant to be a sibling of Y" or whatever.

cguess:
My only question is the text alignment here (Safari)
Screenshot 2022-11-14 at 12 09 56 PM

That may not be worth worrying about though. Otherwise, freakin'great work!

cguess (approved):

reefdog:
@cguess I was able to fix the Safari issue, thanks for noticing it!

PR #460, opened 2022-11-14
Fix video searching

### LINKED ISSUE #459: Media search is returning dupe video results
When I search for a piece of media, Zenodotus returns duplicate results for videos.

To recreate:
1. `rails db:seed` (if you haven't seeded already)
2. Search for a piece of media. I tried [this](https://user-images.githubusercontent.com/9507998/201453232-531f5c47-ab9e-4978-839e-9559e7f171aa.mp4) and [this](https://user-images.githubusercontent.com/9507998/201453259-8c6eba04-64f1-4545-8d42-d756c650b8d0.png)
3. You should see search results like: 
image

The last working commit I see is 8940d070a903236c8fe59f12d58d2b8ba78511ff

cguess:
This dedups the search algorithm and returns the best matching frame. It then follows up with making Facebook and Instagram results properly manage videos vs images.

To test:
- Do a media search with something that's in the archive from either Facebook or instagram, then do the other (to guarantee that there's a result in the listing)
- That's about it, nothing will appear if there's an error because the JS throws a 500, if things do appear it works.

closes #459

### DIFF
--- a/app/models/image_search.rb
+++ b/app/models/image_search.rb
@@ -62,7 +62,6 @@ class ImageSearch < ApplicationRecord
       images
     else
       # For videos we have to loop
-
       video_hashes = []
       self.dhashes.each do |dhash|
         video_hashes = video_hashes | Zelkova.graph.search(dhash["dhash"])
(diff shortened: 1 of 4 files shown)


oneroyalace (approved):
LGTM!

PR #466, opened 2022-11-15
Paginate MediaVault dashboard

### LINKED ISSUE #241: Paginate archive so we're not showing it all
Currently, the only way to browse the vault — besides _searching_, which is different — is to use the index / "Most Recent Archived Items" page. However, that dumps the entire archive to page; since after plugging in MediaReview that is hundreds and hundreds of items long, that's Not Great!

We need to provide a better browse interface, starting with pagination.

reefdog:
This PR finally adds pagination to the MediaVault dashboard, locked to 50 items per page. It also adds generic pagination styling, applied to the generic Pagy markup and so automatically picked up by other places (like on the search history w/in user settings):

Screenshot of pagination showing page 1 active and 29 pages available

**Testing:**
1. If you have fewer than 50 items in your local vault, then either add more (lol) or temporarily modify [`ARCHIVE_ITEMS_PER_PAGE`](https://github.com/TechAndCheck/zenodotus/blob/0c11eb4853c2fe42f7cf94b63e545598c77e06f4/app/controllers/media_vault/archive_controller.rb#L12) to be less than your archive size.
2. Load up the Dashboard and scroll to the bottom.
3. ✅ Pagination should appear and should be usable!

Note: This will be added to search results soon as they are further beautified in #364.

Closes #241

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -14,6 +14,7 @@
 @use "header";
 @use "icons";
 @use "images";
+@use "pagination";
 @use "search_box";
 @use "tables";
 @use "typography";
--- a/app/assets/stylesheets/buttons.scss
+++ b/app/assets/stylesheets/buttons.scss
@@ -80,6 +80,15 @@ elements, but the examples will just use `<button>`.
 		--btn--padding--horizontal: 2em;
 	}
 
+	&.disabled {
+		--btn--color: gray;
+		--btn--background-color: transparent;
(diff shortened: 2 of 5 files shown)


cguess (approved):
This looks great! Only question is if 50 is too many? It's a LOT of scrolling What about 15? It also makes it look like we have a lot of elements archived by making the page count really high versus "3" we have now.

reefdog:
> Only question is if 50 is too many?

Yeah, may be! I'll knock it down to 15.

PR #471, opened 2022-11-16
Add new ClaimReview attributes

### LINKED ISSUE #467: Ingest and handle new ClaimReview fields from FactStream
See https://github.com/TechAndCheck/factstream-backend/commit/1b43047bcc9c8fe40642e2a861b23a310f9e94e4

Steps:
- [x] Update  the `ClaimReview` data model
- [x] Update the ClaimReview JSONSchema file
- [x] Update the FactCheckInsights CSV export
- [x] Update the FactCheckInsights json export
- [x] Update our tests to reflect the new fields

oneroyalace:
This PR adds new sub-attributes to the `ClaimReview` model. One of these attributes, `ClaimReview.itemReviewed.appearance` is an array, so following our prior pattern, we only use its first value when exporting FactCheckInsights data to CSV. 

Closes #467

### DIFF
--- a/app/models/claim_review.rb
+++ b/app/models/claim_review.rb
@@ -17,10 +17,14 @@ class ClaimReview < ApplicationRecord
 
     review_rating "reviewRating_@type" do |review_rating| review_rating["@type"] end
     review_rating "reviewRating_ratingValue" do |review_rating| review_rating["ratingValue"] end
+    review_rating "reviewRating_ratingExplanation" do |review_rating| review_rating["ratingExplanation"] end
     review_rating "reviewRating_bestRating" do |review_rating| review_rating["bestRating"] end
+    review_rating "reviewRating_worstRating" do |review_rating| review_rating["worstRating"] end
     review_rating "reviewRating_image" do |review_rating| review_rating["image"] end
     review_rating "reviewRating_alternateName" do |review_rating| review_rating["alternateName"] end
 
+    item_reviewed "itemReviewed_firstAppearance_@type" do |item_reviewed| item_reviewed.dig("appearance", 0, "@type") end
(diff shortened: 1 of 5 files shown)


app/models/claim_review.rb
```diff
@@ -17,10 +17,14 @@ class ClaimReview < ApplicationRecord
 
     review_rating "reviewRating_@type" do |review_rating| review_rating["@type"] end
     review_rating "reviewRating_ratingValue" do |review_rating| review_rating["ratingValue"] end
+    review_rating "reviewRating_ratingExplanation" do |review_rating| review_rating["ratingExplanation"] end
     review_rating "reviewRating_bestRating" do |review_rating| review_rating["bestRating"] end
+    review_rating "reviewRating_worstRating" do |review_rating| review_rating["worstRating"] end
```
cguess, line 22: One thing to note (should have noticed before) is this single line syntax for blocks can (should) look like this:
`review_rating "reviewRating_ratingExplanation" { |review_rating| review_rating["ratingExplanation"] }`

cguess: I'm not going to say we need to rewrite everything, because it works, but just for the future.

cguess (approved):

app/models/claim_review.rb
oneroyalace: Oh right. Appreciate the callout

oneroyalace: Hm. The line below doesn't work for me @cguess:
```ruby
date_published "datePublished" { |date_published| date_published.strftime("%Y-%m-%d") }
```
Looking at the `comma` gem's wiki, it seems that syntax can only be used if we accept default csv column names.

reefdog: omg, that's annoying. I would've thought this was language-level and not something that had to be explicitly supported by a library. 😵

Issue #473, opened 2022-11-16
Use standard media templates on search results


reefdog:
Our search results currently use old templates and with junk/stand-in data:

<img width="1354" alt="Screenshot of the current MediaVault search results page showing two results with lots of dummy data" src="https://user-images.githubusercontent.com/4731/202219795-9f22d7f5-0083-4f06-8ec1-f0cd0d02e407.png">

We need to adapt them to use the standard Dashboard styling, and add pagination, etc.

Issue #464, opened 2022-11-15
Text search having trouble with some single words


ericaryan:
http://staging.vault.factcheckinsights.org/

Put trump in search box - button flickers but nothing happens on return or click (same for searching just biden)

reefdog:
I accidentally think I figured out what's going on with each of these, from some other work I'm doing.

- `trump` is due to a bug in the template code for Facebook users. This is about to get fixed in a PR I'm doing for #477.
- `biden` is due to a bug in the `FacebookPost` model, where `reactions` is nil so the `number_of_likes` method errors out. This will probably get fixed when I redo search results in #473, although this certainly shouldn't have been causing an error. (We either need a fallback or something.)

PR #478, opened 2022-11-16
Consolidate author controllers

### LINKED ISSUE #477: Consolidate author controllers
We have a lot of unnecessary duplication across the various types of authors, the consequence of which (in the controller/view side) is lots of duplicated code and it being tremendously easy to forget to update something in one place that should be universal.

We should consolidate the following controllers (and their views and tests) into a single `AuthorsController`:

- `FacebookUsersController`
- `InstagramUsersController`
- `TwitterUsersController`
- `YoutubeChannelsController`

reefdog:
This PR consolidates all the various author-type controllers into a single one. This reduces a ton of controller and view redundancy, which was having a real effect as some gnarly bugs were hiding in the discrepancies between templates. (And once I consolidate search results in #473, even more redundancy will be removed.)

It consists of just one commit, but that was pretty necessary since so much was interlinked. Much of the diff is just renaming/moving files, deleting files, and updating routes.

You'll notice that the author URLs are now `/authors/:platform/:id` (e.g., `/authors/twitter/abc123`) rather than something like `/twitter_user/:id` or `/youtube_channel/:id`.

**Testing:**
1. `rails t`
2. `./bin/dev`
3. View the MediaVault Dashboard
4. Drill into users, preferably _every_ user type (twitter, facebook, etc.)
5. Search for results across all user and post types (note that the styling is still old, but nothing should break, except a search for `biden` may trigger the problem raised in #464)

Closes #477


cguess (changes requested):
For Twitter (only I've tried so far) I'm getting this error
```
11:42:12 web.1  | MediaVault::AuthorsController::UnsupportedPlatform (MediaVault::AuthorsController::UnsupportedPlatform):
11:42:12 web.1  |
11:42:12 web.1  | app/controllers/media_vault/authors_controller.rb:41:in `show'
```

cguess (approved):
I love seeing "This file was deleted" so much, works great!

PR #487, opened 2022-12-12
Add a media upload loading indicator

oneroyalace:
This PR creates an animated cog that spins while users upload media to the image search endpoint. 

It also fixes a bug on `master` wherein a user could run media search without uploading any files. 

Closes issue #457

## Testing instructions
1. `rails assets:precompile`
2. Start the Zenodotus server and try searching for media using search bar's explicit button, as well as using drag-and-drop. You should see a loading indicator appear while the media is uploaded.


cguess:
Unfortunately this isn't working and in a few different ways.

All browsers:
- The search field isn't disabled/grayed out while uploading. It should be so that users can see that the field is "activated" or whatever
- It doesn't actually seem to upload the video or image? When I click submit the spinner pops up, spins, but it's never uploaded and nothing happens
- The box shrinks when the spinner shows up. This isn't a big deal, but if possible it'd be best to not redraw the view

Safari: the spinner starts immediately before the "search" button is clicked. May be something with the "return" signal being fired automatically? In chrome you have to click the button first.

The design is really nice though, and I like the way it's going.

cguess (approved):

PR #490, opened 2023-01-03
Add real number to Insight's page

### LINKED ISSUE #421: Count organizations and countries in database
The Insights home page has a stats counter for number of fact-checks, organizations, and countries currently in the database:

image

Number of fact-checks are easy, but we need to make sure we can count orgs and countries.

I don't see this data in the current schema w/in Zeno. Org is probably buried in the [`claim_review.author` JSON](https://github.com/TechAndCheck/zenodotus/blob/c2fcbddd5d2dbb338ef1214e1adb3da2dab7e2ef/db/schema.rb#L80), but will have to be summed. I'm unsure what "country" means; country where the fact-check was published, or where the claim was published, or where the fact-checking org is housed? In any case, that'll need to be queryable in the Zeno db.

For now, I'm going to use static values for those two stats, so this issue is for:

- [ ] Ensure the org/country data is stored and queryable in the database
- [ ] Replace the static org/country counters on the home page with live values

cguess:
This makes the number of organizations in the Insights data base up to date. Number of countries is valid up until November 1, but will have to be manually updated from time to time.

To test:
1. Pull
2. Checkout the home page for insights... that's about it.

Closes #421


oneroyalace (approved):
LGTM

PR #503, opened 2023-01-25
Add MFA

cguess:
This adds webauthn authentication as a required second factor
authentication.

To do so the following are here:
- After setting the password when setting up an account the user is
  prompted to use MFA to validate their account.
- To do the previous we also add `authenticate_user_and_setup!` to
  replace `authenticate_user!` specfically so that it validates MFA is
set up as well everywhere.
- At login the user is not logged in until after MFA is validated
- Added Lottie for animations

Before this is merged into master there will be branches off of it to do
the following:
- [x] Add recovery codes
- [x] Add TOTP for Firefox users since Firefox only implements Webauthn for
   hardware keys, not the passkey standard. This is absurd, but oh well.
  https://github.com/mdp/rotp
- [x] Add the ability to register multiple keys

To Test:
- Go through the entire account request, confirm, authorize system and
  then create a user, set a password and make sure you can register your
MFA. Do this on vault, not insights, to facilitate later testing
- Log into your account with MFA
- Log out, swtich to insights, try to log in again. Should work just
  fine
- Try it in a few different browsers with a hardware key, that should
  work. Passkey is going to be browser locked.


cguess:
If we register a built-in key in one browser we lock them out of other browsers. Hardware keys are immune to this and work just fine across browsers. However, for instance, TouchID only works on Safari.

This is a problem because, for instance, logging in on an iPhone and then on chrome on the computer will send the user into a very frustrating system.

However, we can't just let them add a new key for every new device because... well, an attacker could just grab a different browser and create credentials.

The answer for this is probably a magic email link (super insecure...) or a temp password generated and then moved to a new device, but that's for another step. Before closing this create an issue to address this.

cguess:
Close #517 #516 #8

PR #504, opened 2023-01-26
Add Recovery Codes for User Account

cguess:
Note: This branches off #8, but this should come first. Still in draft mode until the UI and functionality is working.

Enable recovery codes for if the MFA token is lost. This just creates
them and allows checking, but doesn't actually include the UI to check
yet or the actual resetting of the credentials yet.


cguess:
We now have an interface.

